PR #22, opened 2015-02-07
Update DomainEventHelper for not reusing ThreadLocal event

oscarbou:
It can reuse an Event not corresponding to the same "domain action" on a previous phase (which I suppose was the main intention for reusing).

### DIFF
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/DomainEventHelper.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/DomainEventHelper.java
@@ -148,18 +148,19 @@ public class DomainEventHelper {
         }
         try {
             final PropertyDomainEvent<?, ?> event;
-            if(existingEvent != null && phase.isValidatingOrLater()) {
-                event = existingEvent;
-                setEventOldValue(event, oldValue);
-                setEventNewValue(event, newValue);
-            } else {
-                final Object source = ObjectAdapter.Util.unwrap(targetAdapter);
-                final Identifier identifier = identified.getIdentifier();
-                event = newPropertyDomainEvent(eventType, identifier, source, oldValue, newValue);
-            }
+            final Object source = ObjectAdapter.Util.unwrap(targetAdapter);
+            final Identifier identifier = identified.getIdentifier();
+            event = newPropertyDomainEvent(eventType, identifier, source, oldValue, newValue);
             event.setEventPhase(phase);
             event.setPhase(AbstractInteractionEvent.Phase.from(phase));
-            getEventBusService().post(event);
+
+            // Old and New Values are populated only on the VALIDATION Phase and
+            // afterwards.
+            if ((existingEvent != null) && phase.isValidatingOrLater()) {
+                setEventOldValue(event, oldValue);
+                setEventNewValue(event, newValue);
+            }
+            this.getEventBusService().post(event);
             return event;
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


danhaywood:
I'm not sure about this pull request.... seems that a new event is created for each phase.

In my original design I create a new event for hide, and for disable, but then reuse the same event for validate/pre-execute/post-execute.  This opens up the possibility that validation phase could (as a performance optimization) "push" results onto the event, to pass downstream to the later *execute phases.  That seems a good thing to me.

With your redesign, that wouldn't be possible.  

What issue does the new design address?

oscarbou:
Problem is that previous design was not working properly, as it was getting the event from a ThreadLocal that could contain other Event dispatched "in the middle" of another Event's processing (ie, a wrapped property is set in the middle of another Event processing).

So it was reusing an Event that was not corresponding to the same one on previous phases.

As there can be multiple Events simultaneously executing, the only way I envision, if your use case must be supported, is to change from a ThreadLocal single "existingEvent", to a List of them.

HTH,

Oscar 




> El 7/2/2015, a las 12:29, Dan Haywood <notifications@github.com> escribió:
> 
> I'm not sure about this pull request.... seems that a new event is created for each phase.
> 
> In my original design I create a new event for hide, and for disable, but then reuse the same event for validate/pre-execute/post-execute. This opens up the possibility that validation phase could (as a performance optimization) "push" results onto the event, to pass downstream to the later *execute phases. That seems a good thing to me.
> 
> With your redesign, that wouldn't be possible.
> 
> What issue does the new design address?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/apache/isis/pull/22#issuecomment-73360587>.
> 


Óscar Bou Bou
Responsable de Producto
Auditor Jefe de Certificación ISO 27001 en BSI
CISA, CRISC, APMG ISO 20000, ITIL-F

   902 900 231 / 620 267 520
   http://www.twitter.com/oscarbou <http://www.twitter.com/oscarbou>

   http://es.linkedin.com/in/oscarbou <http://es.linkedin.com/in/oscarbou>

   http://www.GesConsultor.com <http://www.gesconsultor.com/> 




Este mensaje y los ficheros anexos son confidenciales. Los mismos contienen información reservada que no puede ser difundida. Si usted ha recibido este correo por error, tenga la amabilidad de eliminarlo de su sistema y avisar al remitente mediante reenvío a su dirección electrónica; no deberá copiar el mensaje ni divulgar su contenido a ninguna persona.
Su dirección de correo electrónico junto a sus datos personales constan en un fichero titularidad de Gesdatos Software, S.L. cuya finalidad es la de mantener el contacto con Ud. Si quiere saber de qué información disponemos de Ud., modificarla, y en su caso, cancelarla, puede hacerlo enviando un escrito al efecto, acompañado de una fotocopia de su D.N.I. a la siguiente dirección: Gesdatos Software, S.L. , Paseo de la Castellana, 153 bajo - 28046 (Madrid), y Avda. Cortes Valencianas num. 50, 1ºC - 46015 (Valencia). Asimismo, es su responsabilidad comprobar que este mensaje o sus archivos adjuntos no contengan virus informáticos, y en caso que los tuvieran eliminarlos.

danhaywood:
OK, understood.   In which case, your design wins.  
If later on we want to re-introduce the concept of reusing events between phases then we can put the extra machinery back in (Lists, etc).
I'll merge this shortly.

oscarbou:
Many thanks, Dan.

PR #24, opened 2015-03-13
ISIS-1083 resetting dependent choice when selecting another referenced choice

erikdehair:
ISIS-1083 resetting dependent choice when selecting another referenced choice


danhaywood:
resolved as https://issues.apache.org/jira/browse/ISIS-1083

PR #27, opened 2015-03-27
ISIS-789 [DN4] 

jdbranham:
ISIS-789

Many changes were made to support DN4. All tests are passing in the fork.


jdbranham:
I had added a neoapp example, but it may not make sense to merge back to master since there is a dependency on module [1] currently only available in source.

[1] https://github.com/jdbranham/isis/blob/master/example/application/neoapp/webapp/pom.xml#L256

danhaywood:
Per neoapp, I've moved that code (plus the Sprint/isis-neo4j module) into isisaddons/isis-app-neoapp.  Haven't actually tried it out yet, but it compiles ok.  

The main PR is merged into the ISIS-789 branch; I intend to merge that into master next week, all being well.

PR #32, opened 2015-07-08
ISIS-1157: Partial implementation proposal

oscarbou:
Tests still pending.

### DIFF
--- a/core/runtime/src/main/java/org/apache/isis/core/runtime/action/invocation/ActionInvocationFacetWithQueryResultsCache.java
+++ b/core/runtime/src/main/java/org/apache/isis/core/runtime/action/invocation/ActionInvocationFacetWithQueryResultsCache.java
@@ -49,6 +49,9 @@ public class ActionInvocationFacetWithQueryResultsCache extends ActionInvocation
             final ObjectAdapter targetAdapter,
             final ObjectAdapter[] argumentAdapters) {
 
+        final CommandContext commandContext = getCommandContext();
+        final Command command = commandContext != null ? commandContext.getCommand() : null;
+
         final ObjectAdapter result = getQueryResultsCache().execute(new Callable<ObjectAdapter>() {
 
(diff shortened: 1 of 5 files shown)


core/runtime/src/main/java/org/apache/isis/core/runtime/action/invocation/ActionInvocationFacetWithQueryResultsCache.java
```diff
@@ -0,0 +1,87 @@
...
+            final ObjectAdapter targetAdapter,
+            final ObjectAdapter[] argumentAdapters) {
+
+        final ObjectAdapter result = getQueryResultsCache().execute(new Callable<ObjectAdapter>() {
+
+            @Override
+            public ObjectAdapter call() throws Exception {
+                return ActionInvocationFacetWithQueryResultsCache.this.underlyingFacet.invoke(owningAction, targetAdapter, argumentAdapters);
+            }
+        }, ActionInvocationFacetWithQueryResultsCache.class, "invoke", owningAction, targetAdapter, targetAdapter);
```
danhaywood: There could be a rare issue with this... the arguments are used to create a cache key, and because you've used the AIFWQRC.class and "invoke" itself, rather than that of the target method that's ultimately called, it's conceivable that caching could occur incorrectly.

eg:
someObject.foo(int, int) -> someObject.bar(int, int)

If both "foo" and "bar" were marked to be cached, then the "foo" cache could be used for "bar".

oscarbou: I will try to reference originally invoked action class and action name
That way, same cache is used when invoked explicitly on the domain logic.

Bostonncity: Ghghjjj

PR #38, opened 2016-03-01
add svg image suffix support

hobrom:
### Changes:
1. Add svg image suffix support to the image cache. 
2. Allow sub-classes of ImageResourceCacheClassPath to access/modify the IMAGE_SUFFICES field.

### Note:
I'm not sure though, whether the web.xml has to be extended accordingly ...
```xml
<web-app ...>
...
<servlet>
    <servlet-name>Resource</servlet-name>
    <servlet-class>org.apache.isis.core.webapp.content.ResourceServlet</servlet-class>
  </servlet>
  <servlet-mapping>
    <servlet-name>Resource</servlet-name>
    <url-pattern>*.svg</url-pattern>
  </servlet-mapping>
...
</web-app>
```
(This would affect the simple-app template and the like.)


danhaywood:
Hi, happy with this PR, however I have no way to merge/close it because it seems that github - for some reason - is reporting it comes from an unknown repository.  
I'm going to apply the change manually; perhaps you could close this PR manually (we don't have permissions to do that ourselves).
Thx
Dan

hobrom:
Thanks Dan!
I'm closing this PR.

PR #40, opened 2016-04-10
ISIS-1364: Declare and implement persistIfNotAlready in RepositorySer…

oscarbou:
…vice and RepositoryServiceDefault


danhaywood:
Oscar and I have been discussing on dev@isis.apache.org; think that this isn't required, because RepositoryService#persist(...) has these semantics already.  Oscar will update the docs to make this clearer, and (I believe) just close this PR.

PR #45, opened 2016-04-29
To fix bug to run in eclipse with *.launch files

rezarahimi:
To fix a bug to run in eclipse with *.launch files:

1. isis-core-webserver was added to dependencies of pom.xml file of webapp

2. in launch files org.apache.isis.WebServer renamed to org.apache.isis.core.webserver.WebServer


danhaywood:
Thanks for taking the time doing this.

In fact, we generate the archetype from the simpleapp application (fyi as per [1])  

So could you make the equivalent changes to simpleapp?  (I can apply this PR or we can discard it... either way it'll be overwritten at the next release).

[1] http://isis.apache.org/guides/cgcom.html#_cgcom_cutting-a-release_releasing-the-archetype_create-the-archetype).

rezarahimi:
I fixed it in the simpleapp folder and I created a new pull request for that:
https://github.com/apache/isis/pull/46

danhaywood:
many thanks, I'll review and merge.

PR #46, opened 2016-04-29
To fix a bug to run in eclipse with *.launch files:

rezarahimi:
To fix a bug to run in eclipse with *.launch files:
1. isis-core-webserver was added to dependencies of pom.xml file of webapp
2. in launch files org.apache.isis.WebServer renamed to org.apache.isis.core.webserver.WebServer


danhaywood:
Thanks for providing this.

In fact, org.apache.isis.WebServer is correct, it just delegates down to org.apache.isis.core.webserver.WebServer.

As for the change for isis-core-webserver, we used to have the pom.xml like this (or very similar), but this caused various issues.   We now hide this dependency behind a profile that only activates within the IDE (IntelliJ).  So, instead of your change, I've done something similar for Eclipse:

https://github.com/apache/isis/commit/2104b7b8a41a1697b0608e85f5786c746b570d39

Thx again
Dan

danhaywood:
could you close this PR ... we have no way of closing PRs that aren't merged in.  Thx

PR #54, opened 2016-09-22
Fix DateTime marshalling in JodaLocalDateValueSemanticsProvider

hobrom:
This implementation of JodaLocalDateTimeValueSemanticsProvider,
given that 
```
    encodingFormatter = ISODateTimeFormat.basicDateTime()
```
assumes that 
```
    date == encodingFormatter.parseLocalDateTime(
                      encodingFormatter.print(date)
    )
```
which is not true (at least for the latest ver. 2.9.4 of joda-time).

Admittedly I provided an ugly workaround, maybe someone finds a more appropriate solution. And maybe this fact has implications on other parts of the framework as well ...


hobrom:
Code snipped that produced an exception:
```
final DateTimeFormatter encodingFormatter = ISODateTimeFormat.basicDateTime();
		
		encodingFormatter.parseLocalDateTime(
				encodingFormatter.print(new org.joda.time.LocalDateTime()));
```

hobrom:
This pull request referenced the wrong source file. So closing.

PR #55, opened 2016-10-01
ISIS-1509: Allow Factory Methods to be defined

oscarbou:


danhaywood:
Hi Oscar,
I didn't merge this in at the time you raised it because as implemented this code allows us to annotate an action as a factory method, but the rest of the framework doesn't actually *do* anything with that annotation.  Maybe that's fine; in writing this comment it occurs to me that you have a custom UI; having this metadata in the metamodel might be of use for your custom UI even it it isn't initially/yet supported in the Wicket or RO viewer. 

But I do have one question: is the idea that factory actions only intended for domain services?

Also... I note that this PR now has conflicts; if you would like to proceed with the PR, perhaps you could rebase it for me?

Thx
Dan

andi-huber:
Oscar, is your proposed factory-method-marker meant to be equivalent to what CDI's @Produces annotation does?
If so, (I believe) we need to take account for that in our current domain-object/-service/-view instantiation/injection mechanism as well. But as we plan to move towards CDI anyway, factory-methods will be supported out of the box. So the question is, can you/we wait for CDI to come?

oscarbou:
Hi Andi. Really interesting question.

An example Domain use case could be the following:

Imagine end users want 2 different ways to instantiate a Book:
- The "fast" one, requiring only to enter a name and an ISBN, the only mandatory properties on that Domain.
- The "complete" one, allowing in addition to the previous ones, to also enter a description, a reference to the Author and more optional data at the very moment you're instantiating the Entity.

Basically we could have something like:

public class BookRepository {
  public Book createBook(final String ISBN, final String name);
  public Book createBookFromAuthor(final String ISBN, final String name, final Author author);
  public Book createBookForArchiving(final String ISBN, final String name, final String description, final Date edition, ...);
  ...
}

With different "create" methods depending on the domain use cases to implement.
Internally, they could reference the "fast" one to delegate base implementation and set additional properties for each derived use case.

From the domain perspective, as far as I would understand it, we would have different Factory Methods from the domain perspective.

AFAIK, in CDI we could use the @Describes annotation with Qualifiers, as there would be different methods that could implement objects of the same Class and we would need to disambiguate.
But on this example they're effectively creating an instance of the same Book class.

Here is the "user" that selects (for "problem solvers") on (or follows for "process followers") each use case one specific implementation.

As current UI generated by Isis is more oriented towards "problem solvers", giving users the different actions, menus, etc. available, if we annotate in some way all methods that from the Domain perspective can be used to instantiate the Book class from the user's perspective, we can, for example, show all them to the user when a reference must be entered of a given class, allowing to select a previously existing instance or a new one.
Exemplyfing it, imagine we show on the modal dialog invoking an action, for each param that can reference a Book, a drop-down menu with all the previous 3 options, giving the user the opportunity to select an existing Book or create a new one right in that very moment.

Perhaps that behaviour (showing the sub-menu with known factory methods) can be the default one for all params in action dialogs, and perhaps it needs to be avoided by something like @Param(showFactoryMethods=false) - being the default value "true" -.


In this case, would you consider these factory methods a "domain" issue rather than a "technical" one - like CDI, where the annotation would inject new instances automatically -?

Despite that, I never thought of using the @Produces annotation for this and think can be a good solution for the most common use case. 
The equivalent could be then a menu for all Qualifiers for the @Produces associated with the Entity class.
 
Sorry for the extension ...

andi-huber:
Hi Oscar, I think I now better understand, what you actually had in mind with these factory-method-markers. As you made the distinction between "domain" and "technical",  I would say that I meant "technical", when referring to CDI's @Produces annotation.

I got confused with the term 'factory', which in my mind is hard-linked to object creation. If I understand you correctly, you would like the UI to identify certain methods that should act as domain-object lookups or domain-object providers.

These lookups/providers would be a new kind of support methods (like 'choices', 'default', 'hide' and similar). A kind that supports not just one particular action, but instead supports all declared actions that have a corresponding parameter type and are annotated with @Param(showFactoryMethods=true). 

Very interesting idea! Not so easy to implement I guess ... nested modal dialogs!

Cheers, Andi

oscarbou:
Hi Andi,

They’re effectively factory methods, creating new instances. Never return existing ones (not lookups).

We have them currently implemented in our custom viewer and allow the user to create a new instance when asking for a reference on an action param (avoiding context switching through a modal over modal dialog) but also to show a “New” button on each entity showing the different Actions annotated with @FactoryMethod that allow to create a new instance (of the given class or any descendant).

Clearer now? I can attach a screenshot.

andi-huber:
Hi Oscar, so these are factories you say. I was thinking of a more general solution for Isis, that could serve even more use-cases (including factories). But I guess I have no concept of how such a solution would actually look like.

Screenshots? Yes sure.

Cheers

oscarbou:
Hi Andi.

The first image shows an action with a param where you can select an existing "Threat Source" or create one of 2 specialized ones ("Adversarial Threat Agent", "Non Adversarial Threat Agent").  
action-param-with-factory-method

When selected on of them, a  modal-over-modal dialog is showed allowing to create the new object.
modal-over-modal

If created, the viewer automatically select it as the current value for the param.

Did it help? Any doubts?

Cheers,

Oscar

andi-huber:
Thanks Oscar, your screenshots definitely helped a lot. For me the missing hint was, that you have these nice little icons besides your parameter editors, that provide the above mentioned factory selection.

Maybe if we collaborate on this idea, we might come up with a nice generic solution for Isis. (You use factories, but one could use lookups as well.) Still I feel that the current 'action-support-method' mechanism in Isis is very powerful. Maybe that's the place we look first ...

Cheers

oscarbou:
Sure Andi!
For me the "hard thing" is the viewer, and you're proficient on that.

What do you mean with generic lookups? 
Can you explain further or give an example?

Thanks!

andi-huber:
By generic solution I mean abstracting from factories to any supporting method that provides the parameter:

Suppose we have an action with parameters (like in your screenshot). Now for a particular parameter we have this support methods like 'defaultXXX' and 'choicesXXX', which are limited in functionality. We could extend on this by telling Isis, that for a particular parameter in our action there is a certain (or are multiple) provider-method-actions (like your factory-methods) that themselves have arguments. 

Just to give an example of a lookup use-case: 
Imagine you have an entity which you want to pass as an action parameter. Using the 'choices' and 'autocomplete' support methods works great, as long as you conveniently find this entity with a single argument search. But as soon as you would need more than one parameter to narrow down your search, Isis currently shows its limits. Here a new kind of support methods would come in handy.

oscarbou:
Aha! So something like custom “domain” finders.

Interesting !!! And find it useful also ... 

Should we ask for more opinions on this on the dev list? 

Please, lead this if you consider

andi-huber:
The issue we are discussing could be labeled 'New Feature: Parameter Providing Support Methods'.

Anyhow for me it's not a 'must have' at the moment. (JEE/CDI support has priority right now.) But it's definitely an extension idea to keep in mind!

Cheers

andi-huber:
It has become unlikely, that this PR will every be merged. So I'm finally closing it.

However, we are well aware of a missing UI feature here!

Any progress regarding this topic is tracked here [1]. Dan has already worked out on 'paper' how this feature could look like, summarized here [2].

[1] https://issues.apache.org/jira/browse/ISIS-1509
[2] https://github.com/andi-huber/isis-2-demo/wiki/Delegates

PR #58, opened 2016-10-17
Remove unsused field declarations

bibryam:


danhaywood:
Thanks, merging using https://issues.apache.org/jira/browse/ISIS-1522

PR #79, opened 2017-05-17
ISIS-1619 - blob/clob client side caching

hobrom:
Regarding https://issues.apache.org/jira/browse/ISIS-1619

If an action is annotated with any of the NON_IDEMPOTENT semantics, client-side caching is now deactivated for Blob/Clob (download) actions.


hobrom:
I need to rebase this, so closing.

PR #80, opened 2017-05-25
Patchset to fix menu actions (ISIS-1625/26)

hobrom:
* ISIS-1625, tooltips for menu actions
* ISIS-1626, confirmation dialog for no-parameter menu actions


hobrom:
merging with PR #79, so closing

PR #84, opened 2017-08-18
ISIS-1616 - support for "As-is" HTML rendering

hobrom:
+ introducing a new value type: Markup
+ Markup holds immutable HTML
+ Markup is applicable as action result and as property type

**Demo snipped:**

    @javax.jdo.annotations.Column(allowsNull = "true")
    @Property(editing = Editing.ENABLED)
    @Getter @Setter
    private String html;
    
    public Markup getMarkup() {
    	return new Markup(html);
    }
    
    @Action(semantics = SemanticsOf.IDEMPOTENT)
    public Markup calcMarkup() {
    	return getMarkup();
    }


hobrom:
Fix for ISIS-1715 menuOrder not honored.
https://issues.apache.org/jira/browse/ISIS-1715

danhaywood:
Thanks for this, Andi, will try to review and get merged soon.  

Quick request: in future, could you raise separate PRs for separate features, rather than adding commits to an existing branch?  I know it's more effort your side, but it helps keep the git history clean.

Thx again
Dan

hobrom:
Thanks Dan, and sorry for the single-branch / multi-feature mix-up ... still learning git. I'll try to do better next time!

BR Andi

PR #88, opened 2017-09-01
Update _ugsec_hints-and-tips_shiro-caching.adoc

ppirus:


danhaywood:
Thanks for this, unfortunately I can't (currently) merge this because you removed your clone. ASF have recently provided the ability to do merges directly through the github UI, so once that's been enabled for our repo, I'll merge in this PR.

ppirus:
again in bold:

-We don't want to repeat the entire link:http://shiro.apache.org/documentation.html[Shiro documentation set] here, but we should flag a number of **_otherof_** other features that are worth checking out.

PR #114, opened 2018-07-11
ISIS-1967: Add tests from Diffblue Deeptest

petersear:
The test cases contained in this PR were automatically generated by Diffblue's Deeptest software.
These tests, and others similarly generated, can be viewed at http://diff.blue/demo.
As discussed with Andi Huber on dev@isis.apache.org, this relatively small selection of tests are provided for assessment and we are prepared to supply further test cases via subsequent PR's if you are happy to receive them.
We would be very interested to hear any feedback you have on these tests as we aim to continuously improve them.


andi-huber:
Thanks Peter!
We are now able to review your submission and discuss internally how to proceed. Hopefully just within a few days, we should be able to give you feedback.

Best regards, Andi

petersear:
Thanks Andi
I really appreciate you taking the time to consider my PR and providing feedback on the tests proposed for inclusion in isis.
We are on the way to resolving some of your concerns:

1. Naming of tests is something we had already identified for improvement and, in fact, the latest analysis includes these changes.
As an example:
```public void org_apache_isis_applib_internal_base__Bytes_prepend_001_8ef33eec105cda50() throws Throwable {```
is now:
```public void prepend1() throws Throwable {```

2. Deeptest software aims to identify all paths through the code and produces a test for each case. The idea is to maximize coverage and remove the possibility that an area of code is not covered. This means a large number of tests will be generated - these are compiled and run by the software with only those which pass being made available by default. It is not suggested that each of these is manually inspected because they represent the current working code and serve to track whether that original behaviour changes in the future (due to a bug or re-design)

3. We do not currently support dependency injection but we are working on it and can let you know when it is available.

Thanks again for your time. We would be happy to provide further tests for your consideration, after the naming and appearance-related changes which have been made recently, and to discuss any issues remaining you may have.

Best regards

Peter

PR #120, opened 2018-09-04
Isis 1976 rethink object adapters

andi-huber:
Key changes so far ...

Interface AdapterManger which allowed read/write access to the ObjectAdapter - cache/map is now gone.

This is replaced by new interface ObjectAdapterProvider, which does not expose the (if any) ObjectAdapter - cache/map.

This affects a lot of facet classes within the metamodel.


core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/concurrency/ConcurrencyChecking.java
```diff
@@ -0,0 +1,104 @@
+package org.apache.isis.core.metamodel.adapter.concurrency;
```
andi-huber: ConcurrencyChecking: missing license header

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaArrayFacet.java
```diff
@@ -19,46 +19,52 @@
...
-    public Collection<ObjectAdapter> collection(final ObjectAdapter collectionAdapter) {
-        final Object[] array = array(collectionAdapter);
+    public Collection<ObjectAdapter> collection(final ObjectAdapter arrayAdapter) {
+        final Object[] array = pojoArray(arrayAdapter);
+        if(isEmpty(array)) {
+            return Collections.emptyList();
+        }
         final ArrayList<ObjectAdapter> objectCollection = new ArrayList<ObjectAdapter>(array.length);
-        for (final Object element2 : array) {
-            final ObjectAdapter element = getAdapterManager().getAdapterFor(element2);
```
andi-huber: change in behavior: 
BEFORE: only a map lookup, which would not trigger ObjectAdapter creation
AFTER: triggers ObjectAdapter creation if required

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaArrayFacet.java
```diff
@@ -67,29 +73,32 @@ public void init(final ObjectAdapter collectionAdapter, final ObjectAdapter[] in
      * Expected to be called with a {@link ObjectAdapter} wrapping an array.
      */
     @Override
-    public ObjectAdapter firstElement(final ObjectAdapter collectionAdapter) {
-        final Object[] array = array(collectionAdapter);
-        return array.length > 0 ? getAdapterManager().getAdapterFor(array[0]) : null;
```
andi-huber: change in behavior:
BEFORE: only a map lookup, which would not trigger ObjectAdapter creation
AFTER: triggers ObjectAdapter creation if required

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaCollectionFacet.java
```diff
@@ -24,28 +24,39 @@
...
+    
+    @Override
+    public Collection<ObjectAdapter> collection(final ObjectAdapter collectionAdapter) {
+        final Collection<?> pojoCollection = pojoCollection(collectionAdapter);
+        
         //TODO [ahuber] java doc states, this is a live view, don't know if this is needed, 
         // or if a copy is sufficient
-        return Collections2.transform(collectionOfUnderlying,
-                ObjectAdapter.Functions.adapter_ForUsing(getAdapterManager()));
+        return Collections2.transform(pojoCollection,
```
andi-huber: Is this live-view actually required?

core/runtime/src/main/java/org/apache/isis/core/runtime/persistence/adapter/PojoAdapter.java
```diff
@@ -123,20 +118,12 @@ public void replacePojo(final Object pojo) {
...
-
-    // -- getOid, replaceOid
+    // -- getOid
     @Override
     public Oid getOid() {
         return oid;
     }
 
-    @Override
-    public void replaceOid(Oid persistedOid) {
```
andi-huber: field in oid in ObjectAdapter is now final
oid replacement no longer allowed

core/runtime/src/main/java/org/apache/isis/core/runtime/services/background/CommandExecutorServiceDefault.java
```diff
@@ -348,44 +349,7 @@ private static OneToOneAssociation findOneToOneAssociationElseNull(
...
-    }
-
-    private ObjectAdapter adapterFor(final Bookmark bookmark) {
-        final RootOid rootOid = RootOid.create(bookmark);
-        return adapterFor(rootOid);
-    }
-
-    private ObjectAdapter adapterFor(final RootOid rootOid) {
-        return getPersistenceSession().adapterFor(rootOid);
+        return ObjectAdapterLegacy.__CommandExecutorServiceDefault.adapterFor(targetObject);
```
andi-huber: refactoring experiment, I could simply revert this change

core/runtime/src/main/java/org/apache/isis/core/runtime/sessiontemplate/AbstractIsisSessionTemplate.java
```diff
@@ -90,46 +78,47 @@ protected void doExecuteWithTransaction(final Object context) {}
...
-
-    protected final ObjectAdapter adapterFor(final Bookmark bookmark) {
-        final RootOid rootOid = RootOid.create(bookmark);
-        return adapterFor(rootOid);
-    }
-
-    protected final ObjectAdapter adapterFor(final RootOid rootOid) {
-        return getPersistenceSession().adapterFor(rootOid);
-    }
+//TODO[ISIS-1976] not used !?
```
andi-huber: possibly changes behavior, need to investigate

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext.java
```diff
@@ -0,0 +1,457 @@
...
+import org.apache.isis.core.metamodel.spec.feature.OneToManyAssociation;
+import org.apache.isis.core.metamodel.specloader.SpecificationLoader;
+import org.apache.isis.core.runtime.memento.Data;
+import org.apache.isis.core.runtime.system.persistence.PersistenceSession;
+
+/**
+ *  
+ * @since 2.0.0-M2
+ */
+public class ObjectAdapterContext {
```
andi-huber: class is a product of refactoring, needs refinement later on; but is sane so far

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext.java
```diff
@@ -0,0 +1,457 @@
...
+    
+    public static interface MementoRecreateObjectSupport {
+        ObjectAdapter recreateObject(ObjectSpecification spec, Oid oid, Data data);
+    }
+    
+    public MementoRecreateObjectSupport mementoSupport() {
+        return mementoSupportMixin;
+    }
+    
+    // ------------------------------------------------------------------------------------------------
```
andi-huber: from here downward: code is work in progress

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_AdapterManager.java
```diff
@@ -0,0 +1,189 @@
...
+import org.apache.isis.core.metamodel.adapter.oid.ParentedCollectionOid;
+import org.apache.isis.core.metamodel.adapter.oid.RootOid;
+import org.apache.isis.core.metamodel.services.ServicesInjector;
+import org.apache.isis.core.runtime.system.persistence.PersistenceSession;
+
+/**
+ *  
+ * @since 2.0.0-M2
+ */
+class ObjectAdapterContext_AdapterManager {
```
andi-huber: package private mixin for ObjectAdapterContext: responsibility AdapterManager 'legacy' support

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_Consistency.java
```diff
@@ -0,0 +1,119 @@
...
+import org.slf4j.LoggerFactory;
+
+import org.apache.isis.core.metamodel.adapter.ObjectAdapter;
+import org.apache.isis.core.metamodel.adapter.oid.Oid;
+
+/**
+ *  
+ * @since 2.0.0-M2
+ */
+class ObjectAdapterContext_Consistency {
```
andi-huber: package private mixin for ObjectAdapterContext: responsibility Cache/Map consistency

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_Factories.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.isis.core.metamodel.specloader.SpecificationLoader;
+import org.apache.isis.core.runtime.persistence.adapter.PojoAdapter;
+import org.apache.isis.core.runtime.system.persistence.PersistenceSession;
+import org.apache.isis.core.runtime.system.persistence.adaptermanager.ObjectAdapterContext.ObjectAdapterFactories;
+
+/**
+ *  
+ * @since 2.0.0-M2
+ */
+class ObjectAdapterContext_Factories implements ObjectAdapterFactories {
```
andi-huber: package private mixin for ObjectAdapterContext: responsibility provides ObjectAdapter factories

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_MementoSupport.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.isis.core.runtime.memento.ObjectData;
+import org.apache.isis.core.runtime.memento.StandaloneData;
+import org.apache.isis.core.runtime.system.persistence.PersistenceSession;
+import org.apache.isis.core.runtime.system.persistence.adaptermanager.ObjectAdapterContext.MementoRecreateObjectSupport;
+
+/**
+ *  
+ * @since 2.0.0-M2
+ */
+class ObjectAdapterContext_MementoSupport implements MementoRecreateObjectSupport {
```
andi-huber: package private mixin for ObjectAdapterContext: responsibility provide object recreation to mementos

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_ObjectAdapterProvider.java
```diff
@@ -0,0 +1,175 @@
...
+import org.apache.isis.core.metamodel.spec.ObjectSpecification;
+import org.apache.isis.core.metamodel.spec.feature.OneToManyAssociation;
+import org.apache.isis.core.metamodel.specloader.SpecificationLoader;
+import org.apache.isis.core.runtime.system.persistence.PersistenceSession;
+
+/**
+ *  
+ * @since 2.0.0-M2
+ */
+class ObjectAdapterContext_ObjectAdapterProvider implements ObjectAdapterProvider {
```
andi-huber: package private mixin for ObjectAdapterContext: responsibility provides ObjectAdapterProvider implementation

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterLegacy.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.isis.schema.common.v1.CollectionDto;
+import org.apache.isis.schema.common.v1.OidDto;
+import org.apache.isis.schema.common.v1.ValueDto;
+import org.apache.isis.schema.common.v1.ValueType;
+import org.apache.isis.schema.utils.CommonDtoUtils;
+
+/**
+ * Interim class, expected to be removed with https://issues.apache.org/jira/browse/ISIS-1976 
+ */
+public class ObjectAdapterLegacy {
```
andi-huber: temporary class for refactoring

core/viewer-wicket-impl/src/main/java/org/apache/isis/viewer/wicket/viewer/integration/wicket/ConverterForObjectAdapter.java
```diff
@@ -44,8 +45,11 @@
      */
     @Override
     public ObjectAdapter convertToObject(final String value, final Locale locale) {
```
andi-huber: Appears not used to me, but not really sure!

core/viewer-wicket-impl/src/main/java/org/apache/isis/viewer/wicket/viewer/integration/wicket/ConverterForObjectAdapterMemento.java
```diff
@@ -54,8 +55,11 @@ public ObjectAdapterMemento convertToObject(final String value, final Locale loc
             return null;
         }
         final Oid oid = RootOid.deStringEncoded(value);
-        final ObjectAdapter adapter = getPersistenceSession().getAdapterFor(oid);
-        return ObjectAdapterMemento.createOrNull(adapter);
+
```
andi-huber: Method appears not used to me, but not really sure!

andi-huber:
Also just discovered a StackOverFlow in the context of the RecreatableObjectFacet, which of course needs fixing:

Call-Cycle
```java
o.a.i.c.runtime.system.persistence.PersistenceSession5.newIdentifier(PersistenceSession5.java:1222)
o.a.i.c.runtime.system.persistence.PersistenceSession5.createTransientOrViewModelOid(PersistenceSession5.java:1194)
o.a.i.c.runtime.system.persistence.adaptermanager.ObjectAdapterContext_ObjectAdapterProvider.adapterFor(ObjectAdapterContext_ObjectAdapterProvider.java:77)
o.a.i.c.metamodel.adapter.ObjectAdapterProvider$Delegating.adapterFor(ObjectAdapterProvider.java:71)
o.a.i.c.metamodel.adapter.ObjectAdapterProvider$Delegating.adapterFor(ObjectAdapterProvider.java:71)
o.a.i.c.metamodel.facets.object.recreatable.RecreatableObjectFacetDeclarativeInitializingAbstract.memento(RecreatableObjectFacetDeclarativeInitializingAbstract.java:115)

o.a.i.c.runtime.system.persistence.PersistenceSession5.newIdentifier(PersistenceSession5.java:1222)
o.a.i.c.runtime.system.persistence.PersistenceSession5.createTransientOrViewModelOid(PersistenceSession5.java:1194)
o.a.i.c.runtime.system.persistence.adaptermanager.ObjectAdapterContext_ObjectAdapterProvider.adapterFor(ObjectAdapterContext_ObjectAdapterProvider.java:77)
o.a.i.c.metamodel.adapter.ObjectAdapterProvider$Delegating.adapterFor(ObjectAdapterProvider.java:71)
o.a.i.c.metamodel.adapter.ObjectAdapterProvider$Delegating.adapterFor(ObjectAdapterProvider.java:71)
o.a.i.c.metamodel.facets.object.recreatable.RecreatableObjectFacetDeclarativeInitializingAbstract.memento(RecreatableObjectFacetDeclarativeInitializingAbstract.java:115)

```

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaArrayFacet.java
```diff
@@ -19,46 +19,52 @@
...
+        final Object[] array = pojoArray(arrayAdapter);
+        if(isEmpty(array)) {
+            return Collections.emptyList();
+        }
         final ArrayList<ObjectAdapter> objectCollection = new ArrayList<ObjectAdapter>(array.length);
-        for (final Object element2 : array) {
-            final ObjectAdapter element = getAdapterManager().getAdapterFor(element2);
-            objectCollection.add(element);
+        for (final Object element : array) {
+            final ObjectAdapter elementAdapter = getObjectAdapterProvider().adapterFor(element);
```
andi-huber: change in behavior:
BEFORE: only a map lookup, which would not trigger ObjectAdapter creation
AFTER: triggers ObjectAdapter creation if required

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaArrayFacet.java
```diff
@@ -67,29 +73,32 @@ public void init(final ObjectAdapter collectionAdapter, final ObjectAdapter[] in
...
     @Override
-    public ObjectAdapter firstElement(final ObjectAdapter collectionAdapter) {
-        final Object[] array = array(collectionAdapter);
-        return array.length > 0 ? getAdapterManager().getAdapterFor(array[0]) : null;
+    public ObjectAdapter firstElement(final ObjectAdapter arrayAdapter) {
+        final Object[] array = pojoArray(arrayAdapter);
+        if(isEmpty(array)) {
+            return null;
+        }
+        return array.length > 0 ? getObjectAdapterProvider().adapterFor(array[0]) : null;
```
andi-huber: change in behavior:
BEFORE: only a map lookup, which would not trigger ObjectAdapter creation
AFTER: triggers ObjectAdapter creation if required

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaCollectionFacet.java
andi-huber: no change in behavior, but ...
Is this live-view actually required?

core/runtime/src/main/java/org/apache/isis/core/runtime/persistence/adapter/PojoAdapter.java
danhaywood: +1  This looks like it's the main significant change in this PR... with the replaceOid stuff now removed.

core/runtime/src/main/java/org/apache/isis/core/runtime/sessiontemplate/AbstractIsisSessionTemplate.java
danhaywood: AbstractIsisSessionTemplate is public API, so while it might not be used within the framework, it's possible that it's used by application code.  However, let's remove it for now, see if we can simplify.  If we do need an API like this, it should probably take Bookmark (from the applib) rather than RootOid.  Also, should probably return just an Object (the domain object) rather than an ObjectAdapter - ie let's keep ObjectAdapter as an internal implementation detail and not in any of our APIs.

core/viewer-wicket-impl/src/main/java/org/apache/isis/viewer/wicket/viewer/integration/wicket/ConverterForObjectAdapterMemento.java
danhaywood: Not sure either. 

IConverter<T> is part of Wicket's SPI, to convert strings that are input in forms to underlying objects.  So this one seems to convert a string representation of an object (eg "cust.Customer:123", like a Bookmark#toString() would return) into an ObjectAdapterMemento, which is basically the Wicket UI's serializable model for an ObjectAdapter (more or less the same as a RootOid).

So, my guess is that it isn't used, because I can't think of anywhere that we have this abiliity for a user to enter "cust.Customer:123" and have the framework look up the corresponding object reference.

Therefore, let's not worry about it too much.  Perhaps add a TODO to see about removing this entire class.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaArrayFacet.java
danhaywood: I think this is ok.

I presume that if  element == null then this implies that elementAdapter == null ?

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaArrayFacet.java
danhaywood: I think this is ok.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaCollectionFacet.java
danhaywood: It's probably would be fine to be a copy, but is there any need to change the behaviour?

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext.java
```diff
@@ -0,0 +1,470 @@
...
+
+        try {
+            pojoAdapterMap.close();
+        } catch(final Throwable ex) {
+            // ignore
+            ObjectAdapterLegacy.LOG.error("close: pojoAdapterMap#close() failed; continuing to avoid memory leakage");
+        }
+    }
+    
+    // -- CACHING
```
danhaywood: I see we do still have our caches, for the moment at least.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
```diff
@@ -0,0 +1,107 @@
...
+            final ObjectAdapter parentAdapter,
+            OneToManyAssociation collection);
+
+    /**
+     * @param viewModelPojo
+     * @return an ObjectAdapter 'bypassing mapping', that holds the ObjectSpecification
+     * FIXME[ISIS-1976] Note: whether or not 'bypassing mapping' should not be exposed by the API.
+     * So this further needs refactoring. 
+     */
+    ObjectAdapter specificationForViewModel(final Object viewModelPojo);
```
danhaywood: This method name seems strange, as compared to the "adapterForViewModel" method below, because it returns an ObjectAdapter, not an ObjectSpecification.

My guess of what you're doing here is that the returned ObjectAdapter is ok to call "getSpecification()" on, but is in some sense incomplete because it has bypassed mapping?

Having these "incomplete" ObjectAdapters around seems dangerous though.  Your comment does indeed have a FIXME.  Is this code used elsewhere? What calls it?

danhaywood:
I think this PR is fine to merge.  

To me it seems that mostly this is just moving functionality about ... the OidAdapterMap and PojoAdapterMap are still present, but relocated to ObjectAdapterContext.

The most significant change is that ObjectAdapter#rootOid is now final, and so the replacePojo stuff has gone, instead the ObjectAdapter is copied if the rootOid changes.  This seems ok to me - or at least will be once we get to the stage where ObjectAdapters are instantiated as required and then discarded immediately afterwards.

The ObjectAdapterProvider#specificationForViewModel(...) - which returns an ObjectAdapter - looks a bit odd; I've created a separate comment for that.  I suspect this is work in progress so happy to see how it plays out, but this code does look weird at the moment.

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSessionBase.java
```diff
@@ -134,16 +128,12 @@ protected PersistenceSessionBase(
...
         // sub-components
-        final AdapterManager adapterManager = this;
-        this.persistenceQueryFactory = new PersistenceQueryFactory(adapterManager, this.specificationLoader);
+        this.persistenceQueryFactory = new PersistenceQueryFactory(
+                obj->this.getObjectAdapterProvider().adapterFor(obj), 
+                this.specificationLoader);
         this.transactionManager = new IsisTransactionManager(this, /*authenticationSession,*/ servicesInjector);
 
         this.state = State.NOT_INITIALIZED;
-
```
danhaywood: Did this configuration property get removed?  Or was the code just moved elsewhere?  I'd prefer to keep it than lose it.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaArrayFacet.java
andi-huber: Yes, pojo is nullable in ObjectAdapterContext_ObjectAdapterProvider ...
```java
public ObjectAdapter adapterFor(Object pojo) {
        
        if(pojo == null) {
            return null;
        }
...
}
```

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/collections/javautilcollection/JavaCollectionFacet.java
andi-huber: no need to change behavior yet; just a standing question, that prevents removal of guava's class Collections2

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext.java
andi-huber: yes, the point of merging this current state is to reduce the scope of java classes that are involved with coming changes with ISIS-1976

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
andi-huber: The original code, had me fooled, because it exploited a side-effect. So I had to quick fix this late last night. 
https://github.com/apache/isis/commit/2c7041016faf87443befaa80eef9a0457ac1338f

The interface for ObjectAdapterProvider will definitely change with further refactoring step.

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSessionBase.java
andi-huber: regarding property 'concurrencyCheckingGloballyEnabled': code moved here

https://github.com/apache/isis/blob/2c7041016faf87443befaa80eef9a0457ac1338f/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/concurrency/ConcurrencyChecking.java#L32

PR #121, opened 2018-09-07
Isis 1976 rethink object adapters

andi-huber:
A http web request might trigger multiple `PersistenceSession` life-cycles (at least one).

Previously each PersistenceSession created its own ObjectAdapter list for any registered domain-service eagerly, even if not required. We do this now lazy:

1) `ObjectAdapterProvider` provides a `List<ObjectAdapter> getServices()` which provides this list on demand, chached in the context of the current PersistenceSession.
2) `ObjectAdapterContext` provides a lookup `ObjectAdapter lookupServiceAdapterFor(RootOid rootOid)` while an underlying lookup map is created lazy and cached application scoped.

Other than that ObjectAdapters are immutable now. (field for the pojo is final)

I'd like to merge at this state, because it appears stable.

### DIFF
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
@@ -487,18 +487,18 @@ public interface ObjectAdapter extends Instance {
      */
     ObjectAdapter withPojo(Object newPojo);
 
-    @Deprecated
-    public static interface Friend {
-      /**
-      * Sometimes it is necessary to manage the replacement of the underlying
-      * domain object (by another component such as an object store). This method
-      * allows the adapter to be kept while the domain object is replaced.
-      */
-     void replacePojo(Object pojo);
-    }
-    
-    @Deprecated
-    Friend friend();
+//    @Deprecated
+//    public static interface Friend {
+//      /**
+//      * Sometimes it is necessary to manage the replacement of the underlying
+//      * domain object (by another component such as an object store). This method
+//      * allows the adapter to be kept while the domain object is replaced.
+//      */
+//     void replacePojo(Object pojo);
+//    }
+//    
+//    @Deprecated
+//    Friend friend();
 
 
 }
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
@@ -469,19 +469,6 @@ public interface ObjectAdapter extends Instance {
             return Util::unwrap;
         }
 
-        public static Function<Object, ObjectAdapter> adapterForUsing(final ObjectAdapterProvider adapterProvider) {
-            return adapterProvider::adapterFor;
-        }
-
-        @Deprecated
-        public static com.google.common.base.Function<Object, ObjectAdapter> adapter_ForUsing(final ObjectAdapterProvider adapterProvider) {
-            return new com.google.common.base.Function<Object, ObjectAdapter>() {
-                @Override
-                public ObjectAdapter apply(final Object pojo) {
-                    return adapterProvider.adapterFor(pojo);
-                }
-            };
-        }
     }
 
     /**
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
@@ -108,13 +108,6 @@ public interface ObjectAdapter extends Instance {
     Instance getInstance(Specification specification);
 
     /**
-     * Sometimes it is necessary to manage the replacement of the underlying
-     * domain object (by another component such as an object store). This method
-     * allows the adapter to be kept while the domain object is replaced.
-     */
-    void replacePojo(Object pojo);
-
-    /**
      * For (stand-alone) collections, returns the element type.
      *
      * <p>
@@ -168,12 +161,6 @@ public interface ObjectAdapter extends Instance {
      * are not mapped and have a <tt>null</tt> oid.
      */
     Oid getOid();
-
-//    /**
-//     * Since {@link Oid}s are now immutable, it is the reference from the
-//     * {@link ObjectAdapter} to its {@link Oid} that must now be updated.
-//     */
-//    void replaceOid(Oid persistedOid);
 
     /**
      * Returns either itself (if this is a root) or the parented collections, the
@@ -499,11 +486,32 @@ public interface ObjectAdapter extends Instance {
 
     /**
      * 
-     * @param persistedRootOid
+     * @param newOid
      * @return a copy of this adapter, having a new RootOid 
      * @since 2.0.0-M2
      */
     ObjectAdapter withOid(RootOid newOid);
+
+    /**
+     * 
+     * @param newPojo
+     * @return a copy of this adapter, having a new Pojo 
+     * @since 2.0.0-M2
+     */
+    ObjectAdapter withPojo(Object newPojo);
+
+    @Deprecated
+    public static interface Friend {
+      /**
+      * Sometimes it is necessary to manage the replacement of the underlying
+      * domain object (by another component such as an object store). This method
+      * allows the adapter to be kept while the domain object is replaced.
+      */
+     void replacePojo(Object pojo);
+    }
+    
+    @Deprecated
+    Friend friend();
 
 
 }
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
@@ -37,11 +37,11 @@ public interface ObjectAdapterProvider {
     
     // -- INTERFACE
 
-    /**
-     * @param pojo
-     * @return oid for the given domain object 
-     */
-    Oid oidFor(Object domainObject);
+//    /**
+//     * @param pojo
+//     * @return oid for the given domain object 
+//     */
+//    Oid oidFor(Object domainObject);
     
     /**
      * @return standalone (value) or root adapter
@@ -86,10 +86,10 @@ public interface ObjectAdapterProvider {
         @Programmatic
         ObjectAdapterProvider getObjectAdapterProvider();
         
-        @Programmatic
-        default Oid oidFor(Object domainObject) {
-            return getObjectAdapterProvider().oidFor(domainObject);
-        }
+//        @Programmatic
+//        default Oid oidFor(Object domainObject) {
+//            return getObjectAdapterProvider().oidFor(domainObject);
+//        }
         
         @Programmatic
         default ObjectAdapter adapterFor(Object domainObject) {
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
@@ -25,6 +25,7 @@ import org.apache.isis.applib.annotation.Programmatic;
 import org.apache.isis.core.metamodel.adapter.oid.Oid;
 import org.apache.isis.core.metamodel.adapter.oid.RootOid;
 import org.apache.isis.core.metamodel.spec.ObjectSpecId;
+import org.apache.isis.core.metamodel.spec.ObjectSpecification;
 import org.apache.isis.core.metamodel.spec.feature.OneToManyAssociation;
 
 /**
@@ -56,12 +57,19 @@ public interface ObjectAdapterProvider {
             OneToManyAssociation collection);
 
     /**
-     * @param viewModelPojo
-     * @return an ObjectAdapter 'bypassing mapping', that holds the ObjectSpecification
-     * FIXME[ISIS-1976] Note: whether or not 'bypassing mapping' should not be exposed by the API.
-     * So this further needs refactoring. 
+     * Returns an ObjectAdapter that holds the ObjectSpecification used for 
+     * interrogating the domain object's metadata. 
+     * <p>
+     * Does _not_ perform dependency injection on the domain object. Also bypasses 
+     * caching (if any), that is each call to this method creates a new unique instance.
+     * </p>
+     * 
+     * @param viewModelPojo domain object
+     * @return  
      */
-    ObjectAdapter specificationForViewModel(final Object viewModelPojo);
+    ObjectAdapter disposableAdapterForViewModel(Object viewModelPojo);
+    
+    ObjectSpecification specificationForViewModel(Object viewModelPojo);
 
     ObjectAdapter adapterForViewModel(
             final Object viewModelPojo, 
@@ -97,7 +105,12 @@ public interface ObjectAdapterProvider {
         }
 
         @Programmatic
-        default ObjectAdapter specificationForViewModel(final Object viewModelPojo) {
+        default ObjectAdapter disposableAdapterForViewModel(final Object viewModelPojo) {
+            return getObjectAdapterProvider().disposableAdapterForViewModel(viewModelPojo);
+        }
+        
+        @Programmatic
+        default ObjectSpecification specificationForViewModel(Object viewModelPojo) {
             return getObjectAdapterProvider().specificationForViewModel(viewModelPojo);
         }
 
@@ -114,6 +127,9 @@ public interface ObjectAdapterProvider {
         }
         
     }
+
+
+   
 
 
     
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.function.Function;
 
 import org.apache.isis.applib.annotation.Programmatic;
+import org.apache.isis.core.metamodel.adapter.oid.Oid;
 import org.apache.isis.core.metamodel.adapter.oid.RootOid;
 import org.apache.isis.core.metamodel.spec.ObjectSpecId;
 import org.apache.isis.core.metamodel.spec.feature.OneToManyAssociation;
@@ -36,6 +37,12 @@ public interface ObjectAdapterProvider {
     // -- INTERFACE
 
     /**
+     * @param pojo
+     * @return oid for the given domain object 
+     */
+    Oid oidFor(Object domainObject);
+    
+    /**
      * @return standalone (value) or root adapter
      */
     ObjectAdapter adapterFor(Object domainObject);
@@ -44,7 +51,7 @@ public interface ObjectAdapterProvider {
      * @return collection adapter.
      */
     ObjectAdapter adapterFor(
-            final Object pojo,
+            final Object domainObject,
             final ObjectAdapter parentAdapter,
             OneToManyAssociation collection);
 
@@ -70,6 +77,11 @@ public interface ObjectAdapterProvider {
         
         @Programmatic
         ObjectAdapterProvider getObjectAdapterProvider();
+        
+        @Programmatic
+        default Oid oidFor(Object domainObject) {
+            return getObjectAdapterProvider().oidFor(domainObject);
+        }
         
         @Programmatic
         default ObjectAdapter adapterFor(Object domainObject) {
@@ -102,6 +114,9 @@ public interface ObjectAdapterProvider {
         }
         
     }
+
+
+    
     
 
 }
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanPrimitiveValueSemanticsProvider.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanPrimitiveValueSemanticsProvider.java
@@ -56,21 +56,21 @@ public class BooleanPrimitiveValueSemanticsProvider extends BooleanValueSemantic
     // BooleanValueFacet impl
     // //////////////////////////////////////////////////////////////////
 
-    @Override
-    public void reset(final ObjectAdapter object) {
-        object.replacePojo(Boolean.FALSE);
-    }
-
-    @Override
-    public void set(final ObjectAdapter object) {
-        object.replacePojo(Boolean.TRUE);
-    }
-
-    @Override
-    public void toggle(final ObjectAdapter object) {
-        final boolean current = ((Boolean) object.getObject()).booleanValue();
-        final boolean toggled = !current;
-        object.replacePojo(Boolean.valueOf(toggled));
-    }
+//    @Override
+//    public void reset(final ObjectAdapter object) {
+//        object.replacePojo(Boolean.FALSE);
+//    }
+//
+//    @Override
+//    public void set(final ObjectAdapter object) {
+//        object.replacePojo(Boolean.TRUE);
+//    }
+//
+//    @Override
+//    public void toggle(final ObjectAdapter object) {
+//        final boolean current = ((Boolean) object.getObject()).booleanValue();
+//        final boolean toggled = !current;
+//        object.replacePojo(Boolean.valueOf(toggled));
+//    }
 
 }
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanValueFacet.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanValueFacet.java
@@ -26,9 +26,9 @@ public interface BooleanValueFacet extends Facet {
 
     boolean isSet(ObjectAdapter object);
 
-    void set(ObjectAdapter object);
-
-    void reset(ObjectAdapter object);
-
-    void toggle(ObjectAdapter object);
+//    void set(ObjectAdapter object);
+//
+//    void reset(ObjectAdapter object);
+//
+//    void toggle(ObjectAdapter object);
 }
--- a/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanWrapperValueSemanticsProvider.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanWrapperValueSemanticsProvider.java
@@ -46,26 +46,26 @@ public class BooleanWrapperValueSemanticsProvider extends BooleanValueSemanticsP
     // BooleanValueFacet impl
     // //////////////////////////////////////////////////////////////////
 
-    @Override
-    public void reset(final ObjectAdapter adapter) {
-        adapter.replacePojo(Boolean.FALSE);
-    }
-
-    @Override
-    public void set(final ObjectAdapter adapter) {
-        adapter.replacePojo(Boolean.TRUE);
-    }
-
-    @Override
-    public void toggle(final ObjectAdapter adapter) {
-        final Object currentObj = adapter.getObject();
-        if (currentObj == null) {
-            set(adapter);
-            return;
-        }
-        final boolean current = ((Boolean) currentObj).booleanValue();
-        final boolean toggled = !current;
-        adapter.replacePojo(Boolean.valueOf(toggled));
-    }
+//    @Override
+//    public void reset(final ObjectAdapter adapter) {
+//        adapter.replacePojo(Boolean.FALSE);
+//    }
+//
+//    @Override
+//    public void set(final ObjectAdapter adapter) {
+//        adapter.replacePojo(Boolean.TRUE);
+//    }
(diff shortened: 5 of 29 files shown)


core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
```diff
@@ -44,17 +52,24 @@
...
+     * <p>
+     * Does _not_ perform dependency injection on the domain object. Also bypasses 
+     * caching (if any), that is each call to this method creates a new unique instance.
+     * </p>
+     * 
+     * @param viewModelPojo domain object
+     * @return  
      */
-    ObjectAdapter specificationForViewModel(final Object viewModelPojo);
+    ObjectAdapter disposableAdapterForViewModel(Object viewModelPojo);
```
andi-huber: Certainly not final API yet. But we are getting better.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
```diff
@@ -71,6 +86,11 @@ ObjectAdapter adapterForViewModel(
         @Programmatic
         ObjectAdapterProvider getObjectAdapterProvider();
         
+//        @Programmatic
+//        default Oid oidFor(Object domainObject) {
+//            return getObjectAdapterProvider().oidFor(domainObject);
+//        }
+
```
andi-huber: seemed a nice addition, but did not work at my first attempt of implementing it; commented out for now

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
```diff
@@ -169,12 +162,6 @@
      */
     Oid getOid();
 
-//    /**
-//     * Since {@link Oid}s are now immutable, it is the reference from the
-//     * {@link ObjectAdapter} to its {@link Oid} that must now be updated.
-//     */
-//    void replaceOid(Oid persistedOid);
-
```
danhaywood: could this be deleted rather than just commented out?

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
```diff
@@ -482,28 +469,36 @@ private Functions(){}
...
-                    return adapterProvider.adapterFor(pojo);
-                }
-            };
-        }
     }
 
     /**
      * 
-     * @param persistedRootOid
+     * @param newOid
```
danhaywood: could you add some comments as to why we (still) need to in effect update the RootOid of an adapter.

My guess is that this is to do with the "disposableAdapter" idea - can only call this method on a disposable adapter in order to clone it and create a "real" adapter?

If that's the case, then maybe the direction of travel is that eventually the "disposable adapter" and "regular adapters" will separate into two separate subtypes of a common supertype.  The common supertype will provide access to specifications etc, the "disposable adapter" has this "withOid(..)" as a factor for "regular adapters", and the "getOid()" method is only exposed by the regular adapters.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/services/ServicesInjector.java
```diff
@@ -78,7 +82,7 @@
      * services that are assignable to the type.  It's possible that this is an empty list.
      */
     private final Map<Class<?>, List<Object>> servicesAssignableToType = _Maps.newHashMap();
-    private final Map<Class<?>, Object> serviceByConcreteType = _Maps.newHashMap();
+    private final _Lazy<Map<Class<?>, Object>> serviceByConcreteType = _Lazy.of(this::initServiceByConcreteType);
```
andi-huber: internal API: _Lazy
this is a perfect showoff on how easy it is to replace a final field with a lazy 'wrapper'
I've slightly adopted this idea from 'vavr' former 'java-slang'

(there's also a thread-safe variant to this)

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
```diff
@@ -482,28 +469,36 @@ private Functions(){}
...
+//      * Sometimes it is necessary to manage the replacement of the underlying
+//      * domain object (by another component such as an object store). This method
+//      * allows the adapter to be kept while the domain object is replaced.
+//      */
+//     void replacePojo(Object pojo);
+//    }
+//    
+//    @Deprecated
+//    Friend friend();
+
```
danhaywood: Could this now be deleted

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
```diff
@@ -35,6 +37,12 @@
     
     // -- INTERFACE
 
+//    /**
+//     * @param pojo
+//     * @return oid for the given domain object 
+//     */
+//    Oid oidFor(Object domainObject);
+
```
danhaywood: delete?

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
```diff
@@ -44,17 +52,24 @@
      * @return collection adapter.
      */
     ObjectAdapter adapterFor(
-            final Object pojo,
+            final Object domainObject,
```
danhaywood: +1 !

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
danhaywood: Better name, becoming more obvious what's going on here.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/services/ServicesInjector.java
```diff
@@ -145,16 +149,19 @@ private ServicesInjector(
...
+    public boolean isRegisteredServiceInstance(final Object pojo) {
+        if(pojo==null) {
+            return false;
+        }
+        final Class<?> key = pojo.getClass();
+        final Object serviceInstance = serviceByConcreteType.get().get(key);
+        return Objects.equals(pojo, serviceInstance);
+    }
+    
+
```
andi-huber: isRegisteredServiceInstance: introduced to ease refactoring; currently not used

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
danhaywood: remove

core/metamodel/src/main/java/org/apache/isis/core/metamodel/services/ServicesInjector.java
```diff
@@ -442,6 +456,17 @@ private static void addAssignableTo(final Class<?> type, final List<Object> cand
...
+    
+    private Map<Class<?>, Object> initServiceByConcreteType(){
+        final Map<Class<?>, Object> map = _Maps.newHashMap();
+        for (Object service : services) {
+            final Class<?> concreteType = service.getClass();
+            map.put(concreteType, service);
+        }
+        return map;
+    }
+
```
andi-huber: that's the value supplier for the above _Lazy

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanPrimitiveValueSemanticsProvider.java
```diff
@@ -56,21 +56,21 @@ public ObjectAdapter getDefault(final ObjectAdapter inObject) {
...
+//    public void set(final ObjectAdapter object) {
+//        object.replacePojo(Boolean.TRUE);
+//    }
+//
+//    @Override
+//    public void toggle(final ObjectAdapter object) {
+//        final boolean current = ((Boolean) object.getObject()).booleanValue();
+//        final boolean toggled = !current;
+//        object.replacePojo(Boolean.valueOf(toggled));
+//    }
```
danhaywood: delete these lines?  (I presume you commented them out because they aren't used anywhere?)

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanValueFacet.java
```diff
@@ -26,9 +26,9 @@
...
-    void set(ObjectAdapter object);
-
-    void reset(ObjectAdapter object);
-
-    void toggle(ObjectAdapter object);
+//    void set(ObjectAdapter object);
+//
+//    void reset(ObjectAdapter object);
+//
+//    void toggle(ObjectAdapter object);
```
danhaywood: ditto, delete

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanWrapperValueSemanticsProvider.java
```diff
@@ -46,26 +46,26 @@ public BooleanWrapperValueSemanticsProvider(final FacetHolder holder, final Serv
...
+//    public void toggle(final ObjectAdapter adapter) {
+//        final Object currentObj = adapter.getObject();
+//        if (currentObj == null) {
+//            set(adapter);
+//            return;
+//        }
+//        final boolean current = ((Boolean) currentObj).booleanValue();
+//        final boolean toggled = !current;
+//        adapter.replacePojo(Boolean.valueOf(toggled));
+//    }
```
danhaywood: ditto, delete

core/plugins/jdo-datanucleus-4/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession4.java
```diff
@@ -1171,91 +1173,40 @@ public ObjectAdapter initializeMapAndCheckConcurrency(final Persistable pojo) {
...
-                recreatableObjectFacet != null
-                ? recreatableObjectFacet.memento(pojo)
-                        : newIdentifierFor(pojo, type);
-
-                return newRootId(spec, identifier, type);
-    }
-
-    private String newIdentifierFor(final Object pojo, final Type type) {
-        return type == Type.TRANSIENT
+    public String identifierFor(final Object pojo, final Oid.State type) {
```
andi-huber: replacing Type (which is undecided yet whether to include in public API) with Oid.State which is API

this step can easily be revoked, it seems however, that we only need to keep one of this to enums so that we don't need to have all this 'glue code'

core/metamodel/src/main/java/org/apache/isis/core/metamodel/services/ServicesInjector.java
```diff
@@ -145,16 +149,19 @@ private ServicesInjector(
...
-        }
-        return serviceByConcreteType.containsKey(cls);
+        return serviceByConcreteType.get().containsKey(cls);
     }
 
+    public boolean isRegisteredServiceInstance(final Object pojo) {
+        if(pojo==null) {
+            return false;
+        }
+        final Class<?> key = pojo.getClass();
```
danhaywood: One thing to be aware of ... when we wrap the domain object then the class will be the synthetic class created by ByteBuddy/Javassist, which will be a subclass of the actual class.  This could be an issue.  There is the concept of a ClassSubstitutor, which was introduced to deal with this: if the class is synthetic, then get its superclass.  I think this mechanism (or something equivalent) should be used everywhere.

core/plugins/jdo-datanucleus-4/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession4.java
```diff
@@ -1171,91 +1173,40 @@ public ObjectAdapter initializeMapAndCheckConcurrency(final Persistable pojo) {
...
-        final ObjectSpecId objectSpecId = spec.getSpecId();
-        return new RootOid(objectSpecId, identifier, state);
-    }
-
-    private ObjectSpecification objectSpecFor(final Object pojo) {
-        final Class<?> pojoClass = pojo.getClass();
-        return getSpecificationLoader().loadSpecification(pojoClass);
-    }
-
-
```
andi-huber: moved to ObjectAdapterContext or one of its mixins

core/plugins/jdo-datanucleus-5/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession5_Decouple.java
```diff
@@ -82,7 +82,13 @@ protected PersistenceSession5_Decouple(PersistenceSession5 holder, ObjectAdapter
...
             final RootOid rootOid,
             final ConcurrencyChecking concurrencyChecking) {
-
+        
+        //FIXME[ISIS-1976] guard against service lookup
+        final ObjectAdapter serviceAdapter = objectAdapterContext.lookupServiceAdapterFor(rootOid);
+        if (serviceAdapter != null) {
+            return serviceAdapter;
+        }
+
```
andi-huber: this whole class is regarded temporary ... refectoring in progress

core/plugins/jdo-datanucleus-4/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession4.java
```diff
@@ -1140,15 +1140,17 @@ public ObjectAdapter initializeMapAndCheckConcurrency(final Persistable pojo) {
...
+        if (originalAdapter != null) {
+            ensureRootObject(pojo); //[ahuber] while already mapped has no side-effect
+            originalOid = (RootOid) originalAdapter.getOid();
 
-            final Version originalVersion = adapter.getVersion();
+            final Version originalVersion = originalAdapter.getVersion();
 
             // sync the pojo held by the adapter with that just loaded
-            objectAdapterContext.remapRecreatedPojo(adapter, pojo);
+            newAdapter = objectAdapterContext.remapRecreatedPojo(originalAdapter, pojo);
```
danhaywood: Yeah, ok.   So I think this is equivalent to "merge" in the JPA APIs.

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession.java
```diff
@@ -108,6 +106,14 @@ default ObjectAdapter adapterFor(final RootOid rootOid) {
...
 
     PersistenceManager getPersistenceManager();
+    
+    /**
+     * @param pojo
+     * @param type
+     * @return String representing an object's id.
+     * @since 2.0.0-M2
+     */
+    String identifierFor(Object pojo, State type);
```
andi-huber: as stated above the Oid.State enum or Type enum are exchangable, ultimately only one should remain, the other thrown away;

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_ObjectAdapterProvider.java
```diff
@@ -114,27 +139,71 @@ protected ObjectAdapter addPersistentToCache(final Object pojo) {
...
-            }
-            serviceAdapters.add(serviceAdapter);
-        }
-        return serviceAdapters;
+        return serviceAdapters.get();
     }
     
     // -- HELPER
     
+    private _Lazy<List<ObjectAdapter>> serviceAdapters = _Lazy.of(this::initServiceAdapters);
```
andi-huber: missing 'final' keyword here

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_ServiceLookup.java
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        public Object lookupServiceInstance(RootOid serviceRootOid) {
+            return servicesById.get(serviceRootOid);
+        }
+    }
+    
+    private ServicesByIdResource initLookupResource() {
+        
+        objectAdapterContext.printContextInfo("INIT SERVICE ID LOOKUP RESOURCE");
```
andi-huber: for debugging, expected to be removed

core/plugins/jdo-datanucleus-5/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession5_Decouple.java
```diff
@@ -82,7 +82,13 @@ protected PersistenceSession5_Decouple(PersistenceSession5 holder, ObjectAdapter
...
+        
+        //FIXME[ISIS-1976] guard against service lookup
+        final ObjectAdapter serviceAdapter = objectAdapterContext.lookupServiceAdapterFor(rootOid);
+        if (serviceAdapter != null) {
+            return serviceAdapter;
+        }
+        
         // attempt to locate adapter for the Oid
         ObjectAdapter adapter = objectAdapterContext.lookupAdapterFor(rootOid);
         if (adapter == null) {
```
danhaywood: Eventually I'm hoping that this code will simplify and then become pluggable.  Assuming that we stop supporting transient pojos, instead this code could iterate over a set of "PersistenceProviders", following the chain of responsibility pattern, where the first PersistenceProvider that recognises the format of the rootOid then takes responsibility for retrieving the corresponding pojo from its persistence store.

In the case of a PersistenceProvider for DN, that means a query to the DB.  
In the case of a PersistenceProvider for view models, it means unmarshalling the state from the oid into the pojo.  (fyi, there's also the optional SPI service, UrlEncodingService or something like, whereby the root oid is a key into some other datastore.  So really my "PersistenceProvider" is a generalization of that concept).

core/plugins/jdo-datanucleus-5/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession5_Decouple.java
```diff
@@ -98,6 +104,14 @@ public ObjectAdapter adapterFor(
...
                 throw ex; // just rethrow
             } catch(RuntimeException ex) {
+                
+                System.err.println("------------------------------------------");
+                System.err.println("rootOid: "+rootOid.enString());
+                
+                ex.printStackTrace();
+                System.err.println("------------------------------------------");
+
+
```
danhaywood: Add a FIXME so we don't forget to remove.  Or, better, use a LOG.debug(...)

core/runtime/src/main/java/org/apache/isis/core/runtime/persistence/adapter/PojoAdapter.java
```diff
@@ -55,7 +57,7 @@
     /**
      * can be {@link #replacePojo(Object) replace}d.
      */
-    private Object pojo;
+    private final Object pojo;
```
danhaywood: +1, I like

core/runtime/src/main/java/org/apache/isis/core/runtime/persistence/adapter/PojoAdapter.java
```diff
@@ -108,15 +109,15 @@ public Object getObject() {
...
-    }
+//    /**
+//     * Sometimes it is necessary to manage the replacement of the underlying
+//     * domain object (by another component such as an object store). This method
+//     * allows the adapter to be kept while the domain object is replaced.
+//     */
+//    @Override
+//    public void replacePojo(final Object pojo) {
+//        this.pojo = pojo;
+//    }
```
danhaywood: delete

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_ObjectAdapterProvider.java
```diff
@@ -114,27 +139,71 @@ protected ObjectAdapter addPersistentToCache(final Object pojo) {
...
+//    }
+    
+//    private Oid persistentOid(final Object pojo) {
+//        if (persistenceSession.getPersistenceManager().getObjectId(pojo) == null) {
+//            return null;
+//        }
+//        final RootOid oid = objectAdapterContext.createPersistentOrViewModelOid(pojo);
+//        return oid;
+//    }
+
```
danhaywood: I guess could delete also

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
danhaywood: ok

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/value/booleans/BooleanPrimitiveValueSemanticsProvider.java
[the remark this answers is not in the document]
danhaywood: yep.  I suspect that quite a bit of ValueSemanticsProvider is redundant.  It was implemented before either of the two current viewers (Wicket and RO), and to be honest they both have their own way of doing some of the responsiblities of VSPs, the parsing/encoding of text into value... in the case of Wicket, using its IConverter<> API, for RO I can't remember.

So, instead of VSPs, we should maybe say its the responsibility of the viewer to obtain values from its UI widgets, and just have the metamodel deal in terms of the actual objects (Strings, booleans, dates etc).

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
andi-huber: yes

core/metamodel/src/main/java/org/apache/isis/core/metamodel/services/ServicesInjector.java
danhaywood: Very elegant!

core/plugins/jdo-datanucleus-4/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession4.java
danhaywood: +1

core/metamodel/src/main/java/org/apache/isis/core/metamodel/services/ServicesInjector.java
danhaywood: +1

core/plugins/jdo-datanucleus-5/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession5_Decouple.java
danhaywood: yeah, I figured.  Fine with me

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
andi-huber: It's not related to the disposable OA idea, I guess. 

There is one remaining reference to changing the pojo in an OA which is from the PersistenceSession implementation, when calling `objectAdapterContext.remapAsPersistent(adapter, persistentOid, this);` in ...

```java
/**
     * Called either when an entity is initially persisted, or when an entity is updated;
     * fires the appropriate lifecycle callback
     *
     * <p>
     * The implementation therefore uses Isis' {@link org.apache.isis.core.metamodel.adapter.oid.Oid#isTransient() oid}
     * to determine which callback to fire.
     */
    @Override
    public void enlistCreatedAndRemapIfRequiredThenInvokeIsisInvokePersistingOrUpdatedCallback(final Persistable pojo) {
        final ObjectAdapter adapter = adapterFor(pojo);

        final RootOid rootOid = (RootOid) adapter.getOid(); // ok since this is for a Persistable

        if (rootOid.isTransient()) {
            // persisting
            final RootOid persistentOid = objectAdapterContext.createPersistentOrViewModelOid(pojo);

            objectAdapterContext.remapAsPersistent(adapter, persistentOid, this);

            CallbackFacet.Util.callCallback(adapter, PersistedCallbackFacet.class);
            postLifecycleEventIfRequired(adapter, PersistedLifecycleEventFacet.class);

            changedObjectsServiceInternal.enlistCreated(adapter);

        } else {
            // updating;
            // the callback and transaction.enlist are done in the preDirty callback
            // (can't be done here, as the enlist requires to capture the 'before' values)
            CallbackFacet.Util.callCallback(adapter, UpdatedCallbackFacet.class);
            postLifecycleEventIfRequired(adapter, UpdatedLifecycleEventFacet.class);
        }

        Version versionIfAny = getVersionIfAny(pojo);
        adapter.setVersion(versionIfAny);
    }
```

andi-huber: regarding the disposable OA idea: we might just uses `Instance` instead of OA which also privides the ObjectSpecification. but needs further investigation

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapter.java
andi-huber: yes

core/metamodel/src/main/java/org/apache/isis/core/metamodel/adapter/ObjectAdapterProvider.java
danhaywood: OK, I understand ... this is code on the way to being introduced, rather than on the way to being removed.

andi-huber: exactly!

core/plugins/jdo-datanucleus-5/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession5_Decouple.java
andi-huber: fixme added

danhaywood:
Not for this PR and maybe not for this issue/ticket either, but I had another thought about ObjectAdapters, which you might want to mull on.

As you know, we also have the ability to wrap domainObjects, used for testing and for emitting events when calling a method action from another class.  One thing we've said we'd like to do (Oscar Bou pushed for this, they do this as a matter of course in their app) is to automatically wrap all domain objects.  The wrapper then ensures that the object's invariants are not violated.  We had ideas of maybe relaxing this rule for calls from "within" the module (however we define a module), but always enforcing it for inter-module calls.

Anyway, it occurs to me that maybe the functionality that's within ObjectAdapter could instead be moved into the wrapper object.  As mentioned in another comment on this PR, the ByteBuddy/Javassist creates a subclass and can mix-in other interfaces for the wrapped object to implement - namely WrappingObject.  This is currently a marker interface, but it could instead have the responsibilities of ObjectAdapter, eg to obtain the oid and the ObjectSpecification.  In other words, the domainObject and the ObjectAdapter would in effect be the same object.  (Actually, since WrappingObject is in the applib, probably we would have a further subtype which is visible only to the runtime, eg "WrappingObjectRuntime")

This would then mean that wherever today we pass around an ObjectAdapter, instead we would pass around a WrappingObjectRuntime.

Anyway, something to mull over.

core/runtime/src/main/java/org/apache/isis/core/runtime/persistence/adapter/PojoAdapter.java
andi-huber: removed

core/plugins/jdo-datanucleus-5/src/main/java/org/apache/isis/core/runtime/system/persistence/PersistenceSession5_Decouple.java
andi-huber: added your comment inline to the source file including a FIXME

core/runtime/src/main/java/org/apache/isis/core/runtime/system/persistence/adaptermanager/ObjectAdapterContext_ObjectAdapterProvider.java
andi-huber: meant as code to come in instead of being remove

andi-huber:
Regarding 'wrapping every domain-object': Seems a very interesting approach to me. I noticed this idea floating around some time, but now you explained it to me in a way that I begin to understand.

I'm closing this PR and merge via push from my local workspace, including the latest changes originating from this conversation.

Thx a lot!

PR #122, opened 2018-09-08
ISIS-1979: Refactor logging in headless runtime and prevent configuri…

blaugold:
…g log4j twice

Without these changes, IsisSystem is setting up log4j which is too late since other classes which
bootstrap Isis require logging to work. Currently, HeadlessWithBootstrappingAbstract configures log4j in
addition to IsisSystem. IsisSystem sets up logging either through logging.properties or a fallback. But
logging.properties does not always exist (integration tests). Without logging.properties the fallback
adds a console appender to the root logger in addition to any configuration in logging-integtest.properties,
resulting in duplicated logs in integration tests.

This commit removes any logging setup from IsisSystem. HeadlessWithBootstrappingAbstract now uses the
IsisLoggingConfigurer which is normally used to set up logging. This requires IsisLoggingConfigurer to
be flexible as to the basename of the configuration file.

Also, some code duplication is refactored into LeveledLogger.

### DIFF
--- a/core/log4j/src/main/java/org/apache/isis/core/runtime/logging/IsisLoggingConfigurer.java
+++ b/core/log4j/src/main/java/org/apache/isis/core/runtime/logging/IsisLoggingConfigurer.java
@@ -51,11 +51,15 @@ public class IsisLoggingConfigurer {
      * The root logging level can also be adjusted using command line arguments.
      */
     public void configureLogging(final String configDirectory, final String[] args) {
+        configureLoggingWithFile(configDirectory + "/" + LoggingConstants.LOGGING_CONFIG_FILE, args);
+    }
+
+    public void configureLoggingWithFile(final String configFile, final String[] args) {
         if (loggingSetup) {
             return;
         }
         loggingSetup = true;
-        configureLogging(configDirectory);
+        configureLogging(configFile);
         applyLoggingLevelFromCommandLine(args);
     }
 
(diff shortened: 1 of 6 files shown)


core/runtime/src/main/java/org/apache/isis/core/runtime/headless/logging/LeveledLogger.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.isis.core.runtime.headless.logging;
```
andi-huber: just to let you know: files do need a license header; but I can fix that after merge

core/log4j/src/main/java/org/apache/isis/core/runtime/logging/IsisLoggingConfigurer.java
```diff
@@ -51,11 +51,15 @@ public IsisLoggingConfigurer(Level rootLoggerLevelIfFallback) {
      * The root logging level can also be adjusted using command line arguments.
      */
     public void configureLogging(final String configDirectory, final String[] args) {
+        configureLoggingWithFile(configDirectory + "/" + LoggingConstants.LOGGING_CONFIG_FILE, args);
+    }
+
+    public void configureLoggingWithFile(final String configFile, final String[] args) {
```
andi-huber: +1

andi-huber:
You know, contributions, that reduce lines of code, are the best.
Thanks a lot!

PR #126, opened 2018-09-15
Isis 1976 rethink object adapters

andi-huber:
Key achievements:
- Metamodel: almost all guava Function and Predicate idioms removed
- Metamodel: slight API improvements, such that some of the internal meta-data that is kept in Java
Collections is no longer copied into immutable lists for any calling client, instead Java Streams are returned (which treat the underlying collections un-modifiable by design)
- the ObjectAdapter per-session map/cache is history
- the legacy modules have been removed, for convenince and ease of migration the legacy/transition-1-2- module remains and shall provide legacy glue code
- shiro was moved to plugins, and its artifact renamed, to be inline with the other plugins

PR #127, opened 2018-09-16
ISIS-1976: removes OA map, stream for metamodel, excising guava

danhaywood:
Key achievements:

* Metamodel: almost all guava Function and Predicate idioms removed
* Metamodel: slight API improvements, such that some of the internal meta-data that is kept in Java Collections is no longer copied into immutable lists for any calling client, instead Java Streams are returned (which treat the underlying collections un-modifiable by design)
* the ObjectAdapter per-session map/cache is history
* the legacy modules have been removed, for convenince and ease of migration the legacy/transition-1-2- module remains and shall provide legacy glue code
* shiro was moved to plugins, and its artifact renamed, to be inline with the other plugins


danhaywood:
This is the same as https://github.com/apache/isis/pull/126, but all commits squashed into one and rebased (actually, I used git cherry-pick) on top of v2.
This means that when applied this PR is effectively a fast-forward merge.

andi-huber:
+1 excellent! Thanks a lot, Dan!

PR #138, opened 2019-05-10
[ISIS-2117] Replace instances of Math.random with Random.nextDouble

bd2019us:
[ISIS-2117](https://issues.apache.org/jira/browse/ISIS-2117) Replace Math.random with Random.nextDouble. There is a performance overhead cost when using Math.random over Random.nextDouble. Even moreso, the Random class can allow for easier random manipulation in the future if needed.

### DIFF
--- a/core/metamodel/src/main/java/org/apache/isis/core/runtime/authentication/standard/RandomCodeGenerator10Chars.java
+++ b/core/metamodel/src/main/java/org/apache/isis/core/runtime/authentication/standard/RandomCodeGenerator10Chars.java
@@ -19,6 +19,8 @@
 
 package org.apache.isis.core.runtime.authentication.standard;
 
+import java.util.Random;
+
 public class RandomCodeGenerator10Chars implements RandomCodeGenerator {
 
     private static final int NUMBER_CHARACTERS = 10;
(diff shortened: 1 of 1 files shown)


andi-huber:
I can see 2 enhancements to this class
1) let 'random' be a final field
2) use Random.nextInt(int bound) instead, to improve code readability

I'm using this inspiration to update the class in the 'v2' development branch. Thx!

```java
    private static final int NUMBER_CHARACTERS = 10;
    private static final String CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890";

    private final Random random = new Random(); 
    
    @Override
    public String generateRandomCode() {
        final StringBuilder buf = new StringBuilder(NUMBER_CHARACTERS);
        for (int i = 0; i < NUMBER_CHARACTERS; i++) {
            final int pos = random.nextInt(CHARACTERS.length()); 
            buf.append(CHARACTERS.charAt(pos));
        }
        return buf.toString();
    }
```

andi-huber:
Optimization applied to '2112_Spring' branch, which is designated to be merged into the final 2.0.0 release.

PR #155, opened 2019-11-07
ISIS-2167: first cut github action

danhaywood:
for now, just runs:

mvn -B package --file pom.xml

Let's see if it works, at all.


danhaywood:
abandoning.  It was branched from master (not suprisingly), think we need to change this.

PR #169, opened 2019-12-05
remove empty lines

lamber-ken:
remove empty lines


andi-huber:
Thanks for spotting these!

PR #178, opened 2020-06-23
Secman now sends a UserCreatedEvent

bkalbfus:
The real work of adding default roles is in the client application.  I committed this after testing my client application receiving and using this new event to add default roles to delegated users.

### DIFF
--- a/extensions/security/secman/api/src/main/java/org/apache/isis/extensions/secman/events/UserCreatedEvent.java
+++ b/extensions/security/secman/api/src/main/java/org/apache/isis/extensions/secman/events/UserCreatedEvent.java
@@ -1,0 +1,16 @@
+package org.apache.isis.extensions.secman.events;
+
+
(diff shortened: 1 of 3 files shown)


andi-huber:
Nice!

andi-huber:
@bkalbfus Please note I did some minor changes after this merge, which however will break  your code! Hope you don't mind:

1) renaming the package and slightly refactoring the new **UserCreatedEvent**
2) moving the autoEnableIfDelegatedAndAuthenticated config option from **IsisConfiguration** to SecMan's **SecurityModuleConfig**

for details see commit https://github.com/apache/isis/commit/564e28ccf1e9855b6527b52bfe6d85982abb673d

andi-huber:
I suspect, there should be a builder for the **SecurityModuleConfig** in your App's 'manifest' where you can set it up programmatically.

PR #180, opened 2020-07-18
JAXB Gradle plugin implementation

jdbranham:
Adds XJC task for jaxb in shema project.
Also, in a few projects, adds non-java resources from `src/main/java` to build folder [gradle] as they are excluded by default.


andi-huber:
Thanks a lot for doing this work!

PR #274, opened 2020-11-09
Isis 2454

joerg-rade:
Please review CorsFilterTest: 
* How can the CORS filter added to the filter chain? 
* Is there a way to have uid/pw checked?


regressiontests/stable/src/test/java/org/apache/isis/testdomain/rest/CorsFilterTest.java
```diff
@@ -0,0 +1,75 @@
...
+        Configuration_usingRoSpec.class,
+        IsisModuleViewerRestfulObjectsJaxrsResteasy4.class,
+})
+class CorsFilterTest {
+
+    @LocalServerPort
+    private int port;
+
+    @Test
+    public void ensureCorsFilterIsInChain() {
```
danhaywood, line 62: I don't see how the test body relates to the test name.

Also, could you add //given, //when, //then comments/regions, so that it's clear which is the main behaviour being exercised (in the //when) ?

regressiontests/stable/src/test/java/org/apache/isis/testdomain/rest/CorsFilterTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ActiveProfiles("test")
+@RunWith(SpringRunner.class)
+@SpringBootTest(
+        classes = RestEndpointService.class,
+        properties = {
+//                "sven = pass, admin_role",
```
danhaywood, line 46: should these be removed if they are commented out?  Or perhaps a more useful comment if their absence is notable?

extensions/vro/cors/impl/pom.xml
```diff
@@ -46,6 +46,31 @@
...
+			<artifactId>javax.servlet-api</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.tomcat.embed</groupId>
+			<artifactId>tomcat-embed-core</artifactId>
+			<version>9.0.37</version>
+			<scope>test</scope>
+		</dependency>
+
```
andi-huber, line 73: Might be opinionated, but lets rather not bring in this heavy lifting test infrastructure into these tiny extension modules. That's what the regressiontests module is for.

regressiontests/stable/src/main/java/org/apache/isis/testdomain/conf/Configuration_headlessButSecure.java
```diff
@@ -0,0 +1,153 @@
...
+@Import({
+        IsisModuleCoreRuntimeServices.class,
+        //IsisModuleCoreSecurity.class,
+        IsisModuleSecurityBypass.class,
+        IsisModuleExtModelAnnotation.class, // @Model support
+        IsisModuleExtCorsImpl.class,
+        Configuration_headlessButSecure.HeadlessCommandSupport.class,
+        KVStoreForTesting.class, // Helper for JUnit Tests
+})
+public class Configuration_headlessButSecure {
```
andi-huber, line 65: Do we need to introduce this code duplication here with respect to the existing `Configuration_headless` class? If something is preventing us from reusing `Configuration_headless`, I'd rather see `Configuration_headless` changed/fixed such that we can avoid code duplication. (This might sound picky, but its easy to add lines of code, and hard to get rid of lines of code.)

andi-huber (commented):
Putting potential expectations into perspective: I cannot answer the questions for this PR without doing some experiments myself, so no promises on giving answers any time soon.

PR #373, opened 2021-02-08
ISIS-2502: working on Exception Recognizers 

andi-huber:
UPDATE: This PR only deals with low level issues. For the actual Exception Recognition stuff a new PR will follow.

(FOLLOW UP)
Provide test cases for:

1. ExceptionRecognizerForDataAlreadyExists
2. ExceptionRecognizerForObjectNotFound
3. ExceptionRecognizerForRelatedDataExists
4. ExceptionRecognizerForUnableToSaveData

---
This change is [Reviewable](https://reviewable.io/reviews/apache/isis/373)


andi-huber:
- [x] JPA: SQLIntegrityConstraintViolationException is not an instance of RuntimeException, so cannot be passed to the built-in ex. translation utilities ... ending up with a too generic DataAccessException
- [ ] JDO: On unique constraint violation, we get a NucleusDataStoreException that sets itself as source (infinite causal chain), so ending up with a too generic DataAccessException

PR #391, opened 2021-02-20
ISIS-1678: replace with ClockService

ErnoViitanen-LUT:
Remove org.apache.isis.applib.clock.Clock, replace with org.apache.isis.applib.services.clock.ClockService

### DIFF
--- a/api/applib/src/main/java/org/apache/isis/applib/services/clock/ClockService.java
+++ b/api/applib/src/main/java/org/apache/isis/applib/services/clock/ClockService.java
@@ -18,7 +18,7 @@
  */
 package org.apache.isis.applib.services.clock;
 
-import org.apache.isis.applib.clock.VirtualClock;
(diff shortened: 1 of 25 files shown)


andi-huber:
I think its save to assume, that https://issues.apache.org/jira/browse/ISIS-1678 can already be considered resolved. The `VirtualClock` was only introduced lately. And has a different use-case within the framework than the previous `Clock` had. I'd rather not move it.

Hope you don't mind if I close this PR without merging. Needless to say its much appreciated seeing PRs and people that want to help getting tasks done. Thumbs up!

andi-huber:
Oh @ErnoViitanen-LUT - and by the way, if thats your first PR ever. Cool stuff, keep going!

danhaywood:
Hi Erno,
and to echo what Andi says - join our Slack channel and have a chat first
about what you'd like to work on
Thanks
Dan

PR #456, opened 2021-03-30
ISIS-2346 - Tab Layout isn't using 100% Width

JuhoKon:
The root width was previously set to 100pc, this caused the TabPanel to not be 100% width even if set to 100%. By setting both the root (RoApp) and TabPanel widths to 100% the Tab-layout now is in fact, 100% width.


andi-huber:
Hi @JuhoKon - thanks for your contribution!

PR #473, opened 2021-04-06
ISIS-2573: entity change tracking: convert interaction scoped to

andi-huber:
- [ ] fix entity change enlisting (currently only triggered initially with JDO's pre-dirty, that is once per entity and transaction)
- [ ] deal with nested transactions (that suspend the current tx)

---
This change is [Reviewable](https://reviewable.io/reviews/apache/isis/473)


andi-huber:
WON'T FIX

PR #478, opened 2021-04-09
ISIS-2063: Entering a string for a joda LocalTime causes an error

joonatan-r:
Without this patch applied, input to a joda LocalTime field causes an error.
This patch fixes the problem by adding a panel to be used by joda LocalTime.


andi-huber:
Hi @joonatan-r  - thanks for that one! 🚀

PR #480, opened 2021-04-11
Isis 2550 : impersonation capabiliity, and ImpersonateMenu for secman

danhaywood:
sketches out the implementation.  Not yet tested (and needs some unit tests)


danhaywood:
@andi-huber  - this is a sketch of user impersonation ("run as") that I had in mind.  Note that it is stateful, using the HttpSession to hold onto.

In theory this impersonation capability could be used and exposed by the core framework, but (for now) I've only surfaced it in the context of secman.

PR #507, opened 2021-04-20
ISIS-2015: Text can be changed on non-editable field, without using edit

jmhupanen:
This PR is part of a mandatory course project. The patch fixes the problem with user being able to edit multiline text field without entering edit mode.


viewers/wicket/ui/src/main/java/org/apache/isis/viewer/wicket/ui/pages/bootstrap-overrides-all-v1.css
```diff
@@ -950,6 +950,10 @@ li.select2-results__option {
     padding-right: 15px;
 }
 
+.multiLineStringPanel .scalarValueInlinePromptLink textarea {
```
andi-huber, line 953: I believe bootstrap-overrides-all-v1.css is no longer used, we might as well delete this file.

viewers/wicket/ui/src/main/java/org/apache/isis/viewer/wicket/ui/pages/bootstrap-overrides-all-v2.css
```diff
@@ -979,6 +979,10 @@ li.select2-results__option {
     padding-right: 15px;
 }
 
+.multiLineStringPanel .scalarValueInlinePromptLink textarea {
+    background-color: white;
```
andi-huber, line 983: It might not be a good idea to hard-code a color value that way, as there are different bootstrap themes one can choose to use. There are also fancy 'dark' themes that would require the 'opposite' color.

viewers/wicket/ui/src/main/java/org/apache/isis/viewer/wicket/ui/components/scalars/ScalarPanelTextFieldAbstract.html
```diff
@@ -86,7 +86,7 @@
     </wicket:fragment>
 
     <wicket:fragment wicket:id="textareaInlinePrompt">
-        <textarea class="form-control input-sm" wicket:id="scalarValue"></textarea>
+        <textarea class="form-control input-sm" wicket:id="scalarValue" readonly="readonly"></textarea>
```
andi-huber, line 89: I believe this is valid, but unfortunately changes the appearance, which we yet have not found a way to fix without a headache considering there are different bootstrap themes a user can choose. I figure the CSS fix you are suggesting works for some themes, but unfortunately not for all of them.

andi-huber (commented):
Good thinking, but not a solution that works for all themes.

andi-huber:
Issue can no longer be reproduced, as we just migrated to Bootstrap 4.

PR #513, opened 2021-04-24
2617 wicketstuff.bs4

mwhesse:


viewers/wicket/ui/src/main/java/org/apache/isis/viewer/wicket/ui/components/bookmarkedpages/BookmarkedPagesPanel.html
```diff
@@ -25,7 +25,7 @@
 	<body>
         <wicket:panel>
             <div class="bookmarkedPagesPanel">
-                <div class="showPanelTab bookmarkRibbon bg-light ml-0"></div>
+                <div class="showPanelTab bookmarkRibbon bg-secondary ml-0"></div>
```
mwhesse, line 28: in BS there was a class called navbar-inverted, but that has been removed

viewers/wicket/ui/src/main/java/org/apache/isis/viewer/wicket/ui/pages/jquery.isis.wicket.viewer.js
```diff
@@ -195,6 +195,37 @@ $(function() {
     });
 */
 
+	/*
+	 * Adapted from https://bootstrap-menu.com/detail-basic-hover.html
+	 * Changes:
+	 * - ignoring mouseleave events if hovering over a popover
+	 */
+	if (window.innerWidth > 992) {
```
mwhesse, line 203: This probably needs some more scrutiny, the event listeners should always be registered, but only react on the right screen size

viewers/wicket/ui/src/main/java/org/apache/isis/viewer/wicket/ui/components/bookmarkedpages/BookmarkedPagesPanel.html
mwhesse: We could invent our own "inverted" background colors, "bg-inverted", but we would have to do so for all themes we are supporting

PR #578, opened 2021-06-03
Isis 2713 and Isis 2712: use namespaces for download metamodel, and fix for all framework-defined features

danhaywood:


danhaywood:
decided I'm gonna split it up into two separate PRs.

PR #587, opened 2021-06-08
ISIS-2723: simplify facet precedence logic

andi-huber:
replace
- Facet#isDerived()
- Facet#isFallback()
- Facet#isAlwaysReplace()
- Facet#get/setUnderlyingFacet(...)
with a simple `Facet#getPrecedence(): Facet.Precedence`


core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/object/hidden/HiddenTypeFacetDerivedFromAuthorization.java
```diff
@@ -29,14 +29,16 @@
...
+extends FacetAbstract
+implements HiddenTypeFacet {
 
     public static Class<? extends Facet> type() {
         return HiddenTypeFacet.class;
     }
 
     public HiddenTypeFacetDerivedFromAuthorization(final FacetHolder holder) {
-        super(type(), holder);
+        super(type(), holder, Precedence.DERIVED);
```
danhaywood, line 41: Although this facet's class name has the word "derived" in it, I'm not sure it automatically means that it's precedence is DERIVED.
The semantics of DERIVED (nb: this is not documented anywhere) is for a facet that is inferred from other facets, but which could be overridden by something that is more specific.
That isn't the case here: the type should always be hidden if all of the members are.
It should probably be named "HiddenTypeFacetInferredFromNoAuthorisationToAnyMember" or something similar.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/actions/notinservicemenu/derived/NotInServiceMenuFacetDerivedFromDomainServiceFacet.java
```diff
@@ -34,7 +34,7 @@
 
     public NotInServiceMenuFacetDerivedFromDomainServiceFacet(
             final NatureOfService natureOfService, final FacetHolder holder) {
-        super(holder);
+        super(holder, Precedence.DERIVED);
```
danhaywood, line 37: as per comment below, not sure this is Derived, because it can never be overridden by something more specific.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/object/hidden/HiddenTypeFacetDerivedFromAuthorization.java
andi-huber: Thanks Dan! With this commit I was just preserving the original behavior. But I guess bugs do surface in the process. I'll change it then.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/actions/notinservicemenu/derived/NotInServiceMenuFacetDerivedFromDomainServiceFacet.java
andi-huber: Same here was just trying to preserve prev. behavior. Thanks for pointing out the issue. Guess its save to revert this commit.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/object/hidden/HiddenTypeFacetDerivedFromAuthorization.java
danhaywood: Sure.  Some thoughts (probably obvious) on the other precedences:
* FALLBACK is lowest priority if there are no other facets for that type, eg infer the NamedFacet from the method name if not specified explicitly using an annotation or layout
* the notion of "alwaysReplace" I can't remember, but I suspect it is very similar to fallback.  
* the notion of an underlying facet is where a higher precedence facet can replace another facet, but keep hold of the underlying facet.  In some cases the replacing facet might wrap but delegate to the underlying, eg to translate the name facet.  If the replacing facet doesn't delegate to the underlying, then there's no real need to retain that reference, but (I wouldn't be surprised) if FacetUtil.addFacet(...) always just chains them anyway regardless.

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/actions/notinservicemenu/derived/NotInServiceMenuFacetDerivedFromDomainServiceFacet.java
danhaywood: Probably best to rename the facet here also, get rid of the word "Derived" and replace with something like "Inferred".

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/object/hidden/HiddenTypeFacetDerivedFromAuthorization.java
andi-huber: Thanks, indeed all correct, I agree.

PR #602, opened 2021-06-11
Isis 2726: minor simplifications and renames

danhaywood:
Moves InteractionTracker up to applib, renames to InteractionLayerTracker
Generalises CommandExecutorServiceDefault so can also switch tht eim
Removes CommandExecutorServiceWithTime (refactor command replay to specify SudoPolicy.SWITCH_USER_AND_TIME)
Removes TickingClockService
merges UserServiceDefault with UserService
merges SudoServiceDefault with SudoService
merges ClockServiceDefault with ClockService
merges QueryResultsCacheDefault with QueryResultsCache


api/applib/src/main/java/org/apache/isis/applib/services/user/UserMemento.java
```diff
@@ -156,7 +157,12 @@ public String title() {
      */
     @Collection
     @CollectionLayout(sequence = "1.4")
-    @Getter @Builder.Default
+    public List<RoleMemento> getRoles() {
+        return roles.toList();
+    }
+
+    @Programmatic
+    @Getter(AccessLevel.PRIVATE) @Builder.Default
```
andi-huber, line 165: aren't these conflicting getters: remove annotation?

PR #658, opened 2021-07-07
ISIS-2793: (doesn't work...) possible fix for keycloak invalidating when impersonating

danhaywood:
... don't cache InteractionContext in AuthenticatedWebSessionForIsis, just delegate to the InteractionLayerTracker


danhaywood:
suerceded by ISIS-2793-v2

PR #725, opened 2021-08-19
Isis 2505 catch up with demo examples

andi-huber:


andi-huber:
we should merge more often

joerg-rade:
done via cli

PR #741, opened 2021-09-06
Isis 2867 : @MemberSupport, local mixins for domain services; also 2 new arch tests.

danhaywood:


andi-huber:
Hi @danhaywood - declaring `act` and `disableAct`  in `SwaggerServiceMenu.openRestApi` should fix the build.

PR #769, opened 2021-11-10
Added @ComponentScan to IsisModuleSubdomainsDocxApplib

pdeenen:
so Docx beans are actually registered in the Spring application context


andi-huber:
Hi @pdeenen - thanks for reporting this! I actually already picked up your _Jira_ issue and did provide a slightly different solution, which also spawned some follow up tasks for cleaning up our isis/submodule packages.

PR #817, opened 2021-12-28
Isis 2878

joerg-rade:
The kroviz jar created via `mvn clean install -DskipTests ` is not the same as the one created via `gradle jar`.
Eventually the use of maven-antrun-plugin could be omitted all together (or replaced by maven-dependency-plugin).

The most elegant way to integrate would be a simple addition of 
```
<dependency>
    <groupId>org.apache.isis.incubator.clients</groupId>
     <artifactId>isis-client-kroviz</artifactId>
     <version>${project.version}</version>
</dependency>
```
to demo/domain/pom.xml - once the jar is in the classpath, resources under /public should be available.

### DIFF
--- a/examples/demo/domain/pom.xml
+++ b/examples/demo/domain/pom.xml
@@ -27,29 +27,63 @@
 	<properties>
 	</properties>
 
-	<build>
-		<resources>
-			<resource>
-				<filtering>true</filtering>
-				<directory>src/main/resources</directory>
-				<includes>
-					<include>application.yml</include>
-				</includes>
-			</resource>
-			<resource>
-				<filtering>false</filtering>
-				<directory>src/main/resources</directory>
-			</resource>
-			<resource>
-				<filtering>false</filtering>
-				<directory>src/main/java</directory>
-				<includes>
-					<!-- we include all .java too, so that we can reference it from the descriptions -->
-					<include>**</include>
-				</includes>
-			</resource>
-		</resources>
-	</build>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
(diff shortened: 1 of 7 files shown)


examples/demo/domain/pom.xml
```diff
@@ -27,29 +27,63 @@
...
-				<includes>
-					<!-- we include all .java too, so that we can reference it from the descriptions -->
-					<include>**</include>
-				</includes>
-			</resource>
-		</resources>
-	</build>
+    <build>
+        <plugins>
+            <plugin>
```
andi-huber, line 32: Maybe add a short explanatory comment, for the motivation of adding this plugin. (I guess some copy&past from this PR's description should suffice)

examples/demo/pom.xml
```diff
@@ -238,12 +243,12 @@
 			</activation>
 			<modules>
 				<module>domain</module>
-				<module>javafx</module>
+				<!--module>javafx</module-->
```
andi-huber, line 246: any reason why this was removed?

examples/demo/pom.xml
```diff
@@ -238,12 +243,12 @@
...
 				<module>web</module>
 				<module>wicket/common</module>
 				<module>wicket/jdo</module>
 				<module>wicket/jpa</module>
-				<module>vaadin</module>
+				<!--module>vaadin</module-->
```
andi-huber, line 251: any reason why this was removed?

andi-huber (commented):
Basically looks good to me, I added some comments though.

examples/demo/pom.xml
joerg-rade: couldn't build locally - committed by accident

examples/demo/pom.xml
joerg-rade: couldn't build locally - committed by accident

PR #865, opened 2022-03-03
Fix for ISIS-2967: Definition of maxlength in @Property and @Parameter

pdeenen:
Fix for ISIS-2967: Definition of maxlength in @Property and @Parameter for the @UserLocale in ApplicationUser


andi-huber:
Thanks!

PR #930, opened 2022-05-14
ISIS-3051: metamodel optimizations

andi-huber:
We established ISIS-3051 is not a bug. But while searching for a ghost, I found a few ways to potentially simplify (speed up) MM internals.


andi-huber:
closing, without merge (mostly already cherry-picked into master)

PR #933, opened 2022-05-17
ISIS-3053: Click on no-arg Action with ARE_YOU_SURE Semantics does nothing, when has Tooltip

andi-huber:
Client side Javascript issues: incompatible wicket-stuff-bootstrap-confirm and bootstrap-popover


andi-huber:
Now fixed in 'master'.

PR #946, opened 2022-05-19
ISIS-3059: Bump Spring Boot 2.6.7 -> 2.7.0

andi-huber:
- Purging cloned Spring Graph-QL stuff.
- Seeing glitches with MariaDB based App. But that's just because MariaDB Connector (JDBC) driver version was bumped, can always override.


andi-huber:
Simply purging the clones Spring Graph-QL classe, left us with a 
```
No qualifying bean of type 'org.springframework.graphql.execution.BatchLoaderRegistry' available
```

So we provide a fallback default with `IsisModuleIncViewerGraphqlViewer`.

danhaywood:
Let's merge it, we can fix up GraphQL if needed later.  Would rather not block the upgrade.

PR #1048, opened 2022-08-08
ISIS-3041 rename InteractionFactory to InteractionService

joerg-rade:


andi-huber (approved):
looks good to me - thx!

PR #1074, opened 2022-08-29
ISIS-3193: introduces isDirty flag to Bindable, so that calculating d…

danhaywood:
…ependent defaults don't overwrite


andi-huber:
I think we can model the dirty state in `org.apache.isis.core.metamodel.interactions.managed.ParameterNegotiationModel` instead, by introducing an additional BooleanObservable. Also we should be able modify behavior as intended, without touching any of the Wicket Viewer code. I'll help with that!

andi-huber:
I checked out this branch doing a little bit of refactoring.

PR #1078, opened 2022-09-02
ISIS-3200: broken Value Choices

andi-huber:
- [x] failing regression tests (JDO related)
- [x] failing value showcases (demo app)


persistence/jpa/applib/src/main/java/org/apache/isis/persistence/jpa/applib/integration/IsisEntityListener.java
```diff
@@ -66,23 +63,23 @@
 public class IsisEntityListener {
 
     // not managed by Spring (directly)
-    @Inject private ServiceInjector serviceInjector;
+    //@Inject private ServiceInjector serviceInjector;
```
andi-huber, line 66: optimization: we could store a INJECTONPOINTS_RESOLVED flag into the associated statemanager

andi-huber:
pre-store event fixes
object mementos simplified a lot

wicket viewer 'execution' still broken as was before the branch-off

PR #1079, opened 2022-09-03
ISIS-3202: Executor does not persist changes

andi-huber:


core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/object/entity/EntityFacet.java
```diff
@@ -61,11 +62,16 @@ public String enstringWithCast(final Object primaryKey) {
...
-                        primaryKeyClass.getName()));
+                        primaryKeyClass.getName()))
+            ;
         }
         public T destring(final String stringifiedPrimaryKey) {
             return idStringifier.destring(owningEntityClass, stringifiedPrimaryKey);
         }
+        /** shallow PK detection */
+        public boolean isValid(final Object primaryKey) {
+            return Try.call(()->enstringWithCast(primaryKey)).isSuccess();
```
andi-huber, line 73: refactor this into a check that does not produce stacktraces

PR #1080, opened 2022-09-04
ISIS-3202: Wicket Viewer Executor not persisting on JPA

andi-huber:
This is round 2. We fixed JDO in round 1.


examples/demo/wicket/jpa/src/main/java/demoapp/webapp/wicket/jpa/DemoAppWicketJpa.java
```diff
@@ -89,6 +89,7 @@ public static void main(final String[] args) {
         //IsisPresets.logging(ComponentFactoryRegistryDefault.class, "debug");
         //IsisPresets.logging(EntityModel.class, "debug");
         //IsisPresets.logging(FormExecutorDefault.class, "debug");
+    	System.setProperty("spring.jpa.show-sql", "true");
```
andi-huber, line 92: comment out before merge

core/metamodel/src/main/java/org/apache/isis/core/metamodel/facets/actions/action/invocation/ActionInvocationFacetForDomainEventAbstract.java
```diff
@@ -198,7 +199,10 @@ public Object execute(final ActionInvocation currentExecution) {
                     null);
 
             // the event handlers may have updated the argument themselves
-            val argsAfterEventPolling = updateArguments(initialArgs, actionDomainEvent.getArguments());
+            val argsAfterEventPolling = updateArguments(
+                    owningAction.getParameters(),
```
andi-huber, line 203: does this handle the mixin case as previously?

PR #1134, opened 2022-10-10
ISIS-2209: fix broken contributing link

HeVictor:
This fixes the broken contribution info link for Apache Isis.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -1,3 +1,3 @@
 = Contributing
 
-You can find full details of how to contribute back to Isis at our https://isis.apache.org/conguide/2.0.0-M4/contributing.html[main website].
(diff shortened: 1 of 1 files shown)


andi-huber:
Thanks for spotting that. As this is a built site (using antora) I believe the link should be rather some generic `xref: ...` reference, that antora autoresolves. In other words we should not need site internal static links of sort `https://isis.apache.org/conguide/version-xxx/contributing.html`

danhaywood:
Actually, we are on M8 now, so the patch isn't quite correct.

andi-huber:
@HeVictor could you update the link to 'M8', then I'm happy to merge.

HeVictor:
@andi-huber Apologies for the late response but I wasn't able to get back to this until now, but I see you've made the required change already which is great. Thanks for taking care of it.

andi-huber:
No worries. Thanks for your contribution regardless.

PR #1150, opened 2022-10-20
Rename packages and artifacts

andi-huber:
### Goals

Rename _Java_ packages and _Maven_ artifacts up to the point where all tests succeed and the demo works.

- [x] core
- [x] security
- [x] persistence
- [x] regressiontests
- [x] viewers
- [x] extensions
- [x] valuetypes
- [x] testing
- [x] tooling
- [x] incubator - kroviz
- [x] incubator - graphql
- [x] incubator - commandreplay
- [x] demo

### Non Goals

rename every occurrence of 'isis'


danhaywood:
I guess it's fine to do this rename incrementally, but the isis-nightly build will need a script that can do everything in one go (though can probably ignore regression tests).

andi-huber:
Agreed. I want to do the rename to 'causeway' in a more controlled manner. And also preserve the git history, if possible.

danhaywood:
just updated the goals, there are 2 further modules in the incubator (graphql and commandreplay)

andi-huber:
just artifacts and packages - doing class renaming separately

PR #1226, opened 2022-11-22
Migrate Swagger Model v2 to OpenAPI v3

andi-huber:


danhaywood:
Seems like a good idea, thanks

PR #1365, opened 2023-01-16
CAUSEWAY-3321: Removal of DomainObjectLayout#plural

andi-huber:
- [ ] new use of `HasObjectFeature` has code de-duplication potential
- [ ] purge `NounForms`
- [ ] optimize use of `PropertyOrCollectionAccessorFacetAbstract` (creates ManagedObject, unwraps it, only to be wrapped again by the caller)
- [ ] remove dead code

### DIFF
--- a/core/metamodel/src/main/java/org/apache/causeway/core/metamodel/object/_InternalTitleUtil.java
+++ b/core/metamodel/src/main/java/org/apache/causeway/core/metamodel/object/_InternalTitleUtil.java
@@ -33,25 +33,6 @@ final class _InternalTitleUtil {
 
     // -- TITLE SUPPORT
 
-    String titleString(
-            final String pluralName,
-            final @NonNull TitleRenderRequest titleRenderRequest) {
-
-        val managedObject = titleRenderRequest.getObject();
-
-        if(managedObject.getSpecialization().isUnspecified()) {
-            return managedObject.getTitle();
-        }
-
-        return managedObject.getSpecification().isSingular()
-            ? objectTitleString(titleRenderRequest)
-                    .trim()
-            : collectionTitleString(
-                    pluralName,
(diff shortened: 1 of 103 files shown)


core/metamodel/src/main/java/org/apache/causeway/core/metamodel/object/_InternalTitleUtil.java
```diff
@@ -32,7 +33,9 @@
 
     // -- TITLE SUPPORT
 
-    String titleString(@NonNull final TitleRenderRequest titleRenderRequest) {
+    String titleString(
```
andi-huber, line 36: not used: remove

core/metamodel/src/main/java/org/apache/causeway/core/metamodel/objectmanager/ObjectManager.java
```diff
@@ -66,11 +66,17 @@ default ObjectMemento mementifyElseFail(final @NonNull ManagedObject object) {
                     _Exceptions.unrecoverable("failed to create memento for  %s", object.getSpecification()));
     }
 
-    //TODO why not use loadObject(bookmark) instead
-    ManagedObject demementify(final ObjectMemento memento);
-    //TODO why not use loadObject(bookmark) instead
-    default ManagedObject demementify(final ObjectSpecification spec, final ObjectMemento memento) {
-        return demementify(memento);
+    /**
+     * TODO why not use loadObject(bookmark) instead?
```
andi-huber, line 70: I think: because action parameters have no bookmark

andi-huber:
With this PR there are some interesting observations regarding the metamodel, maybe subject to future work. In the meantime, we ended up with a more straight forward solution.

PR #1370, opened 2023-01-17
ISIS-3327: [Metamodel] Introspection Glitch when overriding Property Getters from Interfaces

andi-huber:


andi-huber:
fixed in master instead

PR #1371, opened 2023-01-17
ISIS-3328: HTML Documentation Service based on the domain model

myroch:
We have created new documentation service which creates the HTML documentation based on the menu and the domain model. Very similar to the sitemap, but in HTML, so the end user can directly show it together with autogenerated description.

I've migrated it into the latest version, I've just hasn't been able to test it. At least it worked fine in ISIS M7.

### DIFF
--- a/api/applib/src/main/java/module-info.java
+++ b/api/applib/src/main/java/module-info.java
@@ -64,6 +64,7 @@ module org.apache.causeway.applib {
     exports org.apache.causeway.applib.services.commanddto;
     exports org.apache.causeway.applib.services.confview;
     exports org.apache.causeway.applib.services.conmap;
+    exports org.apache.causeway.applib.services.documentation;
     exports org.apache.causeway.applib.services.email;
     exports org.apache.causeway.applib.services.error;
     exports org.apache.causeway.applib.services.eventbus;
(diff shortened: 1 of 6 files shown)


andi-huber:
While there is nothing wrong with the code, I don't think we'll merge it into the framework. Mainly because I think the docs created are suitable to provide a documentation blueprint, but in the general case, not enough to be presented to the end-user.

This opinionated of course.

andi-huber:
For now, let's keep this PR open for further discussion and ideas.

danhaywood:
Thanks for this contribution, @myroch .  Andi and I discussed it a bit more, actually we think we will bring this in, we think it's a nice idea.  So, thanks!  @andi-huber  - can you do the honours?

myroch:
Hallo @andi-huber @danhaywood
Thank you very much for the merging. As I said, I wasn't able to test it in the newest version, have you tried then?
Andi, it looks actually very good in our project, we present it to the end-user. I'm curious what is wrong on your side, maybe lack of the description in the annotations? Otherwise we can still fix it.
Anyway thanks guys, my code is mainly based on @yourfrienddhruv 's work, so he should be honored instead.
Regards,
m.

andi-huber:
I'll see that I add this new menu bar entry to our demo app. I might do some tiny refactoring.

And as I said I like the idea, but don't particularly like the programmatic HTML creation code. But as discussed with Dan, we can always improve that later if required.

PR #1410, opened 2023-02-10
CAUSEWAY-3358: Allow SAFE semantics actions to be invoked with only VIEWING permission

andi-huber:
- [x] implement
- [x] add test cases

### DIFF
--- a/core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
+++ b/core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
@@ -63,6 +63,7 @@ import org.apache.causeway.applib.annotation.ActionLayout;
 import org.apache.causeway.applib.annotation.Introspection.IntrospectionPolicy;
 import org.apache.causeway.applib.annotation.LabelPosition;
 import org.apache.causeway.applib.annotation.PromptStyle;
+import org.apache.causeway.applib.annotation.SemanticsOf;
 import org.apache.causeway.applib.annotation.TableDecorator;
 import org.apache.causeway.applib.services.i18n.Mode;
 import org.apache.causeway.applib.services.iactn.Execution;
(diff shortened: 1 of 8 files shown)


core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
```diff
@@ -115,6 +113,14 @@ public CausewayConfiguration(final ConfigurableEnvironment environment) {
     private final Security security = new Security();
     @Data
     public static class Security {
+
+        /**
+         * If set, allows <i>Actions</i> with SAFE Semantics to be invoked with only VIEWING permissions.
```
danhaywood, line 118: trivial : @link to `@Action` annotation and `Semantics` enum?

andi-huber: yup, good point

core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
```diff
@@ -115,6 +113,14 @@ public CausewayConfiguration(final ConfigurableEnvironment environment) {
...
+         * <p>
+         * default: false
+         */
+        private boolean allowActionsWithSafeSemanticsToBeInvokedWithOnlyViewingPermission = false;
```
danhaywood, line 122: perhaps a little bit shorter?  something like:

`actionsWithSafeSemanticsRequireOnlyViewingPermission`

andi-huber: agreed

regressiontests/stable-interact/src/test/java/org/apache/causeway/testdomain/interact/ActionInteractionTest_usingAllowSafeSemantics.java
```diff
@@ -0,0 +1,107 @@
...
+import lombok.val;
+
+@SpringBootTest(
+        classes = {
+                Configuration_headless.class,
+                Configuration_usingInteractionDomain.class,
+                ActionInteractionTest_usingAllowSafeSemantics.AuthorizorDenyUse.class
+        },
+        properties = {
+                "causeway.security.allowActionsWithSafeSemanticsToBeInvokedWithOnlyViewingPermission=TRUE",
```
danhaywood, line 53: Do we need one further test to confirm that when not set, then the behaviour still requires CHANGING perm?

danhaywood:
Thanks for this @andi-huber  ... added a few small comments against the diff.  Otherwise, really nice!

regressiontests/stable-interact/src/test/java/org/apache/causeway/testdomain/interact/ActionInteractionTest_usingAllowSafeSemantics.java
andi-huber: good point, lets add that

PR #1581, opened 2023-04-14
Causeway 3428

joerg-rade:


joerg-rade:
My intention is to merge CAUSEWAY-3428 into spring6.
There should be ~ three files chnanged.

The UI irritates me a bit - please cross check.

andi-huber:
I just merged latest changes from `master` into `spring6`. So this PR is no longer valid and can be closed. I'll close so we don't merge by accident.

PR #1647, opened 2023-05-14
CAUSEWAY-3460: adds config parameter for Wicket viewer

danhaywood:


viewers/wicket/ui/src/main/java/org/apache/causeway/viewer/wicket/ui/components/scalars/ScalarPanelAbstract2.java
```diff
@@ -257,8 +257,10 @@ private void addAdditionalButtonsTo(
                     addDisabledReasonIcon(buttonContainer, "fa-solid fa-ban", "");
                     break;
                 case DISABLED_REASON_PROTOTYPING:
-                    addDisabledReasonIcon(buttonContainer, "fa-solid fa-text-slash icon-prototyping",
-                            " Note: This icon only appears in prototyping mode.");
+                    if (getConfiguration().getViewer().getWicket().isDisableReasonExplanationInPrototypingModeEnabled()) {
```
andi-huber, line 260: I suppose it would better fit to do this check here ... `ScalarPanelAdditionalButton.DISABLED_REASON_PROTOTYPING#isVisible(...)`

danhaywood: good point, will do

PR #1754, opened 2023-06-23
CAUSEWAY-2646: Change layout namespaces bs3->bs

andi-huber:
### Goals

- [x] layout files
- [x] adoc mention

### Non Goals

- rename schema URI


danhaywood:
@andi-huber  ... for the legacy builds, though, can we revert the namespace back to whatever it used to be?

andi-huber:
@danhaywood - I believe what I'm doing here is non invasive. Meaning it should not affect existing code ...

old
```
xmlns:bs3="https://causeway.apache.org/applib/layout/grid/bootstrap3"
```

new
```
xmlns:bs="https://causeway.apache.org/applib/layout/grid/bootstrap3"
```

Agreed? (I might be wrong.)

PR #1755, opened 2023-06-23
CAUSEWAY-3498: skips validation of choices/defaults for always-hidden…

danhaywood:
… actions


core/metamodel/src/main/java/org/apache/causeway/core/metamodel/postprocessors/param/ChoicesAndDefaultsPostProcessor.java
```diff
@@ -68,6 +70,13 @@ public void postProcessParameter(
             final ObjectAction objectAction,
             final ObjectActionParameter param) {
 
+        // no need to check if this action is always hidden in the UI
+        // (eg if it exists solely to be called via the WrapperFactory service)
+        val hiddenFacet = objectAction.getFacet(HiddenFacet.class);
```
andi-huber, line 75: We have convenient utility code for this in `Facets.hiddenWhere(...).`

andi-huber:
I like that one! Also useful for actions that are only exposed via RO, like in `@DomainService(nature=REST)`

PR #1756, opened 2023-06-23
CAUSEWAY-3485: Rename ScalarModel -> PopModel

andi-huber:
PoP .. Property or Parameter


andi-huber:
giving up on this idea

PR #1787, opened 2023-07-12
CAUSEWAY-3522: Entities referenced from within Viewmodels may stay hollow during followup AJAX request

andi-huber:
- [ ] verify that we can accurately track the load/detach lifecycle of viewmodels, that is for both _regular_ and AJAX requests
- [ ] if so, we should be able to remove the wicket viewmodel refresh hackery


andi-huber:
abandoned: instead we fixed this by caching titles within `EntityIconAndTitlePanel`

PR #1843, opened 2023-08-25
Causeway 3557

danhaywood:


commons/src/main/java/org/apache/causeway/commons/collections/Can.java
```diff
@@ -176,29 +188,48 @@ public static <T> Can<T> of(final T ... array) {
...
+     * @param <T>
+     */
+    public static <T> Can<T> ofArray(T[] array, Function<T, T> nullMapper) {
         if(_NullSafe.size(array)==0) {
             return empty();
         }
 
         val nonNullElements = Stream.of(array)
                 .filter(_NullSafe::isPresent)
+                .map(nullMapper)
```
andi-huber, line 219: needs to happen before the filter

commons/src/main/java/org/apache/causeway/commons/collections/Can.java
```diff
@@ -176,29 +188,48 @@ public static <T> Can<T> of(final T ... array) {
...
+    }
 
+    /**
+     *
+     * @param array
+     * @param nullMapper -
+     * @return
+     * @param <T>
+     */
+    public static <T> Can<T> ofArray(T[] array, Function<T, T> nullMapper) {
```
andi-huber, line 212: arguments: **array** is `@Nullable`  while **nullMapper** is `@NonNull`

commons/src/main/java/org/apache/causeway/commons/collections/Can.java
```diff
@@ -59,8 +59,10 @@
  * <p>
  * Same idiomatic convention applies: References to {@link Can}
  * should never be initialized to {@code null}.
+ *
  * <p>
- * A {@link Can} must not contain elements equal to {@code null}.
+ * <b>IMPORTANT:</b>A {@link Can} must not contain elements equal to {@code null}.  If you want to store a null within
```
andi-huber, line 64: I guess the second sentence does not really add any information.

andi-huber: That said, `Can<T>` simply does not store `null`. Working around that fact is bad practice. Instead use a different container.

danhaywood: updated in subsequent commit direct to `master`.

PR #1887, opened 2023-09-21
CAUSEWAY-3597: adds mixin to download .columnOrder.txt file

danhaywood:


core/metamodel/src/main/java/org/apache/causeway/core/metamodel/spec/feature/ObjectAssociation.java
```diff
@@ -157,6 +165,43 @@ public static final Predicate<ObjectAssociation> staticallyVisible(final Where w
...
+
+        public static Predicate<ObjectAssociation> referencesParent(
+                final @Nullable ObjectSpecification parentSpec) {
+            if(parentSpec == null) {
+                return _Predicates.alwaysFalse();
+            }
+            return (final ObjectAssociation assoc) -> {
+                    if(assoc.isCollection()) return false; // never true for collections
+                    return Facets.hiddenWhereMatches(Where.REFERENCES_PARENT::equals).test(assoc)
+                            && parentSpec.isOfType(assoc.getElementType());
```
andi-huber, line 202: Perhaps this needs to be flipped to make more sense:
`assoc.getElementType().isOfType(parentSpec)`

In other words, the match is successful, if the parent type is assignable from the property type. (Not the other way around.)

(I know not a change this PR introduced, as this is just copy pasted code.)

danhaywood: As you noticed, I did just refactor the code, but yes, there's a ticket I raised recently regarding the fact that REFERENCES_PARENT doesn't seem to be honoured.

Actually I think the correct formulation may be assoc.getReturnType().isOfType(parentSpec), because the assoc in question is of the elements of the collection, and we want to exclude those that point back to (have a return type of) the parent owning the collection.

andi-huber: This could support both now, _Properties_ and _Collections_, hence also removing the always-false-for-collections line:
```java
return (final ObjectAssociation assoc) ->
            Facets.hiddenWhereMatches(Where.REFERENCES_PARENT::equals).test(assoc)
                        && assoc.getElementType().isOfType(parentSpec);
```

danhaywood: Thanks, my comment above was incorrect, there is no such method as `assoc.getReturnType()` , the `assoc.getElementType()` is the one.

I think the check to return `false` for Collections is still right, actually.  The purpose of `REFERENCES_PARENT` is for use cases such as `Order` ---> `OrderItem`, where the `OrderItem` refers back to its parent `Order` ; that's redundant information to show in the collection of items if we're viewing the parent `Order`.

Suppose though that `OrderItem` itself had a collection of `Order`s; bit of an odd model, but perhaps something like pointing to previous `Order`s where the product being bought was also purchased.  Anyway, it wouldn't make sense to hide this collection of `OrderItem`, even though its elementType would be `Order`.

andi-huber: I agree with your example, but devs can choose not to use Where.REFERENCES_PARENT in such cases. No?

danhaywood: What I'm saying though is that even if they do accidentally set `hidden=Where.REFERENCES_PARENT`, we should ignore it.  

```
public class OrderItem {
    @Getter
    @CollectionLayout(hidden=Where.REFERENCES_PARENT)   // this is invalid and will be ignored
    private Set<Order> previousOrders;
    ...
}
```

I guess perhaps it should be a MM validation error, not sure I can be bothered; will add some javadoc though.

PR #1903, opened 2023-09-26
CAUSEWAY-3608: adds additional config props to ConfigurationViewModel

danhaywood:


andi-huber:
Root cause of failing CI is
```
Could not resolve placeholder 'LIB' in value "/opt/hostedtoolcache/CodeQL/2.14.5/x64/codeql/tools/linux64/${LIB}_${PLATFORM}_trace.so"
```
Not sure what to make of that yet.

PR #1949, opened 2023-10-12
CAUSEWAY-3630: Simplified Tabular Data Integration

andi-huber:
- [x] `ComponentFactory` implements `Serializable` which is wrong (refactor)
- [x] Decouple export/download extension from Wicket Viewer
- [x] Deprecate ExcelDownload (provide replacement under tabular ext.)
- [x] Allow for multiple export providers to be registered
- [ ] Provide a generic DataTable model for tabular data (powered by the framework's metamodel)


andi-huber:
closing so can reopen against 'master'

PR #2020, opened 2023-11-07
CAUSEWAY-3641: DataTableInteractive#Memento don't require Executions on Deserialization

andi-huber:
- [x] Use the (now) serializable `DataTable` to model table data
- [ ] Invent some freeze mechanism for `DataTable`, that is, once we have the cell data, don't require to recompute (via executions)

As we do want to pave the way for a generic search field, we need all the table rows, hence table paging is modeled as a view into the **full** `DataTable`.


andi-huber:
It is OK to bypass change events for unmarshalling! See discussion https://github.com/apache/causeway/pull/2403

andi-huber:
superseded by https://github.com/apache/causeway/pull/2482

PR #2080, opened 2023-12-07
CAUSEWAY-3655: Layout Inconsitencies

andi-huber:
- [ ] fixup tests (broken since merge of 3579)
- [ ] demo app (jpa) [hidden where](http://localhost:8080/restful/objects/demo.CausewayVegaEntity/88/object-layout) does not agree w/ what Wicket Viewer renders


andi-huber:
Multiple tries, I don't know how to do it. I guess parking this for now.

PR #2259, opened 2024-03-08
CAUSEWAY-3620: throws 404 if object not found ; adds integ test

danhaywood:


andi-huber:
Hi @danhaywood, building on your investigations a one-liner fix would be to just add a filter like
```java
    protected ManagedObject getObjectAdapterElseThrowNotFound(
            final String domainType,
            final String instanceIdEncoded,
            final @NonNull UnaryOperator<RestfulObjectsApplicationException> onRoException) {
        final String instanceIdDecoded = UrlDecoderUtils.urlDecode(instanceIdEncoded);

        val bookmark = Bookmark.forLogicalTypeNameAndIdentifier(domainType, instanceIdDecoded);
        return metaModelContext.getObjectManager().loadObject(bookmark)
                //CAUSEWAY-3620 don't let EMPTY ManagedObject(s) pass
                .filter(_Predicates.not(ManagedObjects::isNullOrUnspecifiedOrEmpty))
                .orElseThrow(()->onRoException.apply(
                        RestfulObjectsApplicationException
                        .createWithMessage(HttpStatusCode.NOT_FOUND,
                                "Could not determine adapter for bookmark: '%s'",
                                bookmark)));
    }
```

andi-huber:
I guess this proposed filter is correct for entities and viewmodels, but not for values, which should be allowed to be `null`. I don't know if this method is ever called for values. (We have bookmark support for values now.)

If in doubt, the fillter should let `ManagedObject`(s) of type VALUE pass.

danhaywood:
Thanks for those suggestions, Andi - I'll refactor that way.  To answer your question, no - this method should never be called for a value, only for entities and view models.

This is an aside, but since you mentioned it: was does a bookmark for a value look like, eg a string "hello world" ?

andi-huber:
> what does a bookmark for a value look like, eg a string "hello world" ?

Each value-type has its ValueTypeSemantics, which defines how to stringify and de-stringify a value. The stringified value is then made URL-SAFE by encoding it as base64 to create the id part of the bookmark.

PR #2403, opened 2024-05-07
CAUSEWAY-3734 excel reader uses pass through interaction

opencirclesolutions-wove:


andi-huber:
Pass-through execution will also ignore any events that are potentially involved with action execution or property change.

Using pass-through for some use cases to speed up things, does make sense to me (e.g. serialization). However, as far as I am concerned, there is that open question regarding the bypassing of events.

danhaywood:
We have our own fork of this module, and so this was the correct solution for us.  It certainly shouldn't be `InteractionInitiatedBy.USER`, our reasoning was that `PASSTHROUGH` (rather than `FRAMEWORK`) was the best fit because in a sense we are unmarshalling a spreadsheet into domain objects.

For me `PASSTHROUGH` seems the safest option and not very controversial.  I suppose if we can't decide then we could provide a config option to specify which, but I think that's unnecessary complexity.

andi-huber:
In other words 'when to allow bypassing of change events' is an open research question, which I must admit, I had not yet taken my time to investigate in more detail. For reference, this is also related to https://github.com/apache/causeway/pull/2020.

andi-huber:
Thanks Dan. I guess you are confirming what I had suspected. Unmarshalling / de-serialization should be a perfect fit with `PASSTHROUGH` execution.

PR #2410, opened 2024-05-16
CAUSEWAY-3741 - Enables hyphen in namespace

bkalbfus:


bkalbfus:
I tested this by building causeway-app-simpleapp jdo-SNAPSHOT branch with the namespace modified to "simple-module".

andi-huber:
Have you considered using `AsciiIdentifierService` SPI instead?

danhaywood:
Hi @andi-huber  - I think that `AsciiIdentifierService` is only called to clean up the memberIds, but not the type names.

image

Given we already have this sanitization, and the code being changed is specific to the graphql viewer only, then I'm happy to have this change here.

danhaywood:
@bkalbfus  - could you add an integration test and optionally a unit test to verify that this new code works correctly?

For example, you could take a copy of the CalculatorService and give it a namespace that requires sanitization?

bkalbfus:
I would gladly!

bkalbfus:
I updated my fork with the tests.  About to make a PR, but that would include commits from this PR.  And, I see my commit is showing on this PR after my comment, so is that enough for you to bring it in?  What is the preferred workflow?

bkalbfus:
I checked what would happen to the tests if I changed the name of the test domain object, `org.apache.causeway.viewer.graphql.viewer.test.domain.i18n.I18nCalculator`, to `university.calc.I18nCälculator`.  A bunch of tests failed, so it looks like this sanitization is justified.  Or, more calls need to be put in to `AsciiIdentifierService`.

andi-huber:
Thanks Brian.

Perhaps my preferred approach would have been to enhance the `AsciiIdentifierService`, in support of sanitizing namespaces. But then I'm also lazy and don't want to do the work. So happy with the current solution as well.

PR #2411, opened 2024-05-17
ESTUP2-668 enable bulk mode for property change audit trails

opencirclesolutions-wove:
When creating new entities the amount of audit trail entries can become quite large and inserting them one by one shouldn’t be necessary when you have a bulk option in the generic repository. The insertions still are being wrapped by another transactional so there is no case of starting a transaction for every insert. It’s just a minor performance increase.


danhaywood:
Tests?

danhaywood:
@opencirclesolutions-wove  - hm, I guess this PR is redundant because the other one includes all the changes here??

danhaywood:
superceded by pr 2412, I believe.

PR #2412, opened 2024-05-17
Estup2-668 adding a bulk mode for the generic repository to enable delaying commits

opencirclesolutions-wove:
When handling large collections it may be beneficial to delay the commit for performance reasons.
This framework is a tread safe way to delay the committing. This only applies of the code is not used within transactional boundries already.

As an example of use the case of audit trail entry insertions in bulk is used.

### DIFF
--- a/api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
+++ b/api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
@@ -19,6 +19,7 @@
 package org.apache.causeway.applib.services.publishing.spi;
 
 import org.apache.causeway.applib.annotation.DomainObject;
+import org.apache.causeway.commons.collections.Can;
 import org.apache.causeway.commons.having.HasEnabling;
 
 /**
@@ -48,4 +49,5 @@ public interface EntityPropertyChangeSubscriber extends HasEnabling {
      */
     void onChanging(EntityPropertyChange entityPropertyChange);
 
+    void onBulkChanging(Can<EntityPropertyChange> entityPropertyChanges);
 }
--- a/api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
+++ b/api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
@@ -49,5 +49,6 @@ public interface EntityPropertyChangeSubscriber extends HasEnabling {
      */
     void onChanging(EntityPropertyChange entityPropertyChange);
 
-    void onBulkChanging(Can<EntityPropertyChange> entityPropertyChanges);
+    default void onBulkChanging(Can<EntityPropertyChange> entityPropertyChanges) {
+    }
 }
--- a/api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
+++ b/api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
@@ -50,5 +50,6 @@ public interface EntityPropertyChangeSubscriber extends HasEnabling {
     void onChanging(EntityPropertyChange entityPropertyChange);
 
     default void onBulkChanging(Can<EntityPropertyChange> entityPropertyChanges) {
+        entityPropertyChanges.forEach(this::onChanging);
     }
 }
--- a/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
+++ b/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
@@ -85,6 +85,14 @@ public interface RepositoryService {
     <T> T detachedEntity(@NonNull T entity);
 
     /**
+     * Suspends flushing transaction for {@param object} instances until the {@param bulkMode} is turned off again.
+     * <p>
+     * Usage should be wrapped in a try {} finally {} construction.
+     *
+     */
+    <T> void setBulkMode(final T object, final Boolean bulkMode);
+
+    /**
      * Persist the specified object (or do nothing if already persistent).
      *
      * <p>
--- a/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
+++ b/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
@@ -85,12 +85,12 @@ public interface RepositoryService {
     <T> T detachedEntity(@NonNull T entity);
 
     /**
-     * Suspends flushing transaction for {@param object} instances until the {@param bulkMode} is turned off again.
+     * Suspends flushing transaction for {@param aClass} instances until the {@param bulkMode} is turned off again.
      * <p>
      * Usage should be wrapped in a try {} finally {} construction.
      *
      */
-    <T> void setBulkMode(final T object, final Boolean bulkMode);
+    <T extends Class> void setBulkMode(final T aClass, final Boolean bulkMode);
 
     /**
      * Persist the specified object (or do nothing if already persistent).
--- a/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
+++ b/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
@@ -20,6 +20,7 @@ package org.apache.causeway.applib.services.repository;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.Callable;
 import java.util.function.Predicate;
 
 import org.springframework.lang.Nullable;
@@ -85,12 +86,12 @@ public interface RepositoryService {
     <T> T detachedEntity(@NonNull T entity);
 
     /**
-     * Suspends flushing transaction for {@param aClass} instances until the {@param bulkMode} is turned off again.
-     * <p>
-     * Usage should be wrapped in a try {} finally {} construction.
+     * Enables bulk mode for all {@param aClass} in {@param callable}.
+     *
+     * Used for a large collection of persist calls without calling flush.
      *
      */
-    <T extends Class> void setBulkMode(final T aClass, final Boolean bulkMode);
+    <T> T execInBulk(Callable<T> callable, Class<?>... classes);
 
     /**
      * Persist the specified object (or do nothing if already persistent).
--- a/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
+++ b/api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
@@ -86,12 +86,12 @@ public interface RepositoryService {
     <T> T detachedEntity(@NonNull T entity);
 
     /**
-     * Enables bulk mode for all {@param aClass} in {@param callable}.
+     * Enables bulk mode for all generic repository calls in {@param callable}.
      *
-     * Used for a large collection of persist calls without calling flush.
+     * Used for a large collection of generic repository calls without calling flush.
      *
      */
-    <T> T execInBulk(Callable<T> callable, Class<?>... classes);
+    <T> T execInBulk(Callable<T> callable);
 
     /**
      * Persist the specified object (or do nothing if already persistent).
--- a/core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
+++ b/core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
@@ -2085,6 +2085,19 @@ public class CausewayConfiguration {
                     Mode mode = Mode.WRITE;
                 }
             }
+
+            private final EntityPropertyChangePublisher entityPropertyChangePublisher = new EntityPropertyChangePublisher();
+
+            @Data
+            public static class EntityPropertyChangePublisher {
+
+                private final Bulk bulk = new Bulk();
+
+                @Data
+                public static class Bulk {
+                    int threshold = 1;
+                }
+            }
         }
     }
 
--- a/core/runtimeservices/src/main/java/org/apache/causeway/core/runtimeservices/publish/EntityPropertyChangePublisherDefault.java
+++ b/core/runtimeservices/src/main/java/org/apache/causeway/core/runtimeservices/publish/EntityPropertyChangePublisherDefault.java
@@ -67,7 +67,7 @@ public class EntityPropertyChangePublisherDefault implements EntityPropertyChang
 
     private Can<EntityPropertyChangeSubscriber> enabledSubscribers = Can.empty();
 
-    @Value("${entity.property.change.publisher.bulk.threshold:10}")
+    @Value("${entity.property.change.publisher.bulk.threshold:1}")
     private int propertyBulkThreshold;
 
     @PostConstruct
--- a/core/runtimeservices/src/main/java/org/apache/causeway/core/runtimeservices/publish/EntityPropertyChangePublisherDefault.java
+++ b/core/runtimeservices/src/main/java/org/apache/causeway/core/runtimeservices/publish/EntityPropertyChangePublisherDefault.java
@@ -27,6 +27,7 @@ import javax.inject.Named;
 import javax.inject.Provider;
 
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.lang.Nullable;
 import org.springframework.stereotype.Service;
 
@@ -66,6 +67,9 @@ public class EntityPropertyChangePublisherDefault implements EntityPropertyChang
 
     private Can<EntityPropertyChangeSubscriber> enabledSubscribers = Can.empty();
 
+    @Value("${entity.property.change.publisher.bulk.threshold:10}")
+    private int propertyBulkThreshold;
+
     @PostConstruct
     public void init() {
         enabledSubscribers = Can.ofCollection(subscribers)
@@ -102,12 +106,17 @@ public class EntityPropertyChangePublisherDefault implements EntityPropertyChang
                     enabledSubscribers,
                     () -> getCannotPublishReason(propertyChanges)
             );
-
-            propertyChanges.forEach(propertyChange->{
-                for (val subscriber : enabledSubscribers) {
-                    subscriber.onChanging(propertyChange);
+            if(propertyChanges.size() <= propertyBulkThreshold) {
+                propertyChanges.forEach(propertyChange -> {
+                    for (val subscriber : enabledSubscribers) {
+                        subscriber.onChanging(propertyChange);
+                    }
+                });
+            } else {
+                for(val subscriber : enabledSubscribers) {
+                    subscriber.onBulkChanging(propertyChanges);
                 }
-            });
+            }
         } finally {
             _Xray.exitPublishing(xrayHandle);
         }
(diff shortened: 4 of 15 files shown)


danhaywood:
@opencirclesolutions-wove  - hm, I guess this PR is redundant because the other one includes all the changes here??

api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
```diff
@@ -48,4 +49,5 @@ public interface EntityPropertyChangeSubscriber extends HasEnabling {
      */
     void onChanging(EntityPropertyChange entityPropertyChange);
 
+    void onBulkChanging(Can<EntityPropertyChange> entityPropertyChanges);
```
danhaywood, line 52: should be `default` for backward compatibliity

extensions/security/audittrail/applib/src/main/java/org/apache/causeway/extensions/audittrail/applib/dom/AuditTrailEntryRepository.java
```diff
@@ -38,6 +39,8 @@ public interface AuditTrailEntryRepository {
 
     AuditTrailEntry createFor(final EntityPropertyChange change);
 
+    Can<AuditTrailEntry> createForBulk(final Can<EntityPropertyChange> entityPropertyChanges);
```
danhaywood, line 42: should be `default` for backward compatibliity

api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
```diff
@@ -84,6 +84,14 @@ public interface RepositoryService {
      */
     <T> T detachedEntity(@NonNull T entity);
 
+    /**
+     * Suspends flushing transaction for {@param aClass} instances until the {@param bulkMode} is turned off again.
+     * <p>
+     * Usage should be wrapped in a try {} finally {} construction.
+     *
+     */
+    <T extends Class> void setBulkMode(final T aClass, final Boolean bulkMode);
```
danhaywood, line 93: Don't like this design, it runs the risk of leaving the service in the 'bulk' mode.  Also, while the impl is on a thread-local, the API suggests that the service is stateful, which is a no-no.

Instead, provide an API that accepts a lambda and does all the work within, eg:

`execInBulk( Callable )`

core/runtimeservices/src/main/java/org/apache/causeway/core/runtimeservices/publish/EntityPropertyChangePublisherDefault.java
```diff
@@ -66,6 +67,9 @@ public class EntityPropertyChangePublisherDefault implements EntityPropertyChang
 
     private Can<EntityPropertyChangeSubscriber> enabledSubscribers = Can.empty();
 
+    @Value("${entity.property.change.publisher.bulk.threshold:1}")
```
danhaywood, line 70: This property name needs to change, and also should be read from `CausewayConfiguration` so that it is typesafe.

The property name should be `causeway.core.runtime-services.entity-property-change-publisher.bulk.threshold`

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
```diff
@@ -116,14 +119,24 @@ public <T> T persist(final T domainObject) {
...
         persist(object);
-        transactionService.flushTransaction();
+        if(!threadLocal.get().getOrDefault(object.getClass(), Boolean.FALSE)) {
+            transactionService.flushTransaction();
+        }
         return object;
     }
 
+    @Override
+    public <T extends Class> void setBulkMode(final T aClass, final Boolean bulkMode) {
```
danhaywood, line 132: See previous comment, instead of this setter we use an `execInBulk(Callable)` action.

danhaywood (changes requested):
In addition, we need some integ tests.  there are integ tests already in the regression suite, suggest add to those.

api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
opencirclesolutions-wove: Fixed

extensions/security/audittrail/applib/src/main/java/org/apache/causeway/extensions/audittrail/applib/dom/AuditTrailEntryRepository.java
opencirclesolutions-wove: Fixed

api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
opencirclesolutions-wove: Fixed

core/runtimeservices/src/main/java/org/apache/causeway/core/runtimeservices/publish/EntityPropertyChangePublisherDefault.java
opencirclesolutions-wove: Fixed

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Fixed

opencirclesolutions-wove:
Test coverage shows that the changes are hit 4 times by regular testing and 22 times by regression tests.

api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
```diff
@@ -48,4 +49,6 @@ public interface EntityPropertyChangeSubscriber extends HasEnabling {
      */
     void onChanging(EntityPropertyChange entityPropertyChange);
 
+    default void onBulkChanging(Can<EntityPropertyChange> entityPropertyChanges) {
```
danhaywood, line 52: Shouldn't the default implementation iterate over the `Can<>` and call `onChanging(...)` ?

api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
```diff
@@ -84,6 +85,14 @@ public interface RepositoryService {
      */
     <T> T detachedEntity(@NonNull T entity);
 
+    /**
+     * Enables bulk mode for all {@param aClass} in {@param callable}.
```
danhaywood, line 89: This javadoc is out of date.

core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
```diff
@@ -2085,6 +2085,19 @@ public static class Po {
...
+            private final EntityPropertyChangePublisher entityPropertyChangePublisher = new EntityPropertyChangePublisher();
+
+            @Data
+            public static class EntityPropertyChangePublisher {
+
+                private final Bulk bulk = new Bulk();
+
+                @Data
+                public static class Bulk {
+                    int threshold = 1;
```
danhaywood, line 2098: We need an integration test that demonstrates when the value of this is changed from its default.

extensions/security/audittrail/applib/src/main/java/org/apache/causeway/extensions/audittrail/applib/dom/AuditTrailEntryRepository.java
```diff
@@ -38,6 +39,10 @@ public interface AuditTrailEntryRepository {
 
     AuditTrailEntry createFor(final EntityPropertyChange change);
 
+    default Can<AuditTrailEntry> createForBulk(final Can<EntityPropertyChange> entityPropertyChanges) {
+        return Can.empty();
```
danhaywood, line 43: I think that the default implementation of this should iterate over the provided `Can<..>` and call the original `createFor(...)`, mapping to the returned `Can<>`.

extensions/security/audittrail/applib/src/main/java/org/apache/causeway/extensions/audittrail/applib/dom/AuditTrailEntryRepositoryAbstract.java
```diff
@@ -60,12 +63,25 @@ public Class<E> getEntityClass() {
...
+    @Override
     public AuditTrailEntry createFor(final EntityPropertyChange change) {
         E entry = factoryService.detachedEntity(auditTrailEntryClass);
         entry.init(change);
         return repositoryService.persistAndFlush(entry);
     }
 
+    @Override
+    public Can<AuditTrailEntry> createForBulk(Can<EntityPropertyChange> entityPropertyChanges) {
+        Collection<AuditTrailEntry> auditTrailEntries = repositoryService.execInBulk(() -> entityPropertyChanges.stream()
```
danhaywood, line 75: If the definition of the supertype's `default` impl is the same as the callable, ie:

```
return entityPropertyChanges.stream()
                .map(change -> {
                    E entry = factoryService.detachedEntity(auditTrailEntryClass);
                    entry.init(change);
                    return repositoryService.persist(entry);
                }).collect(Collectors.toList());
```

then it would then allow this override to be simplified to something like:

```
return Can.ofCollection(
  repositoryService.execInBulk( () -> super.createForBulk(entityPropertyChanges) )
  );
```

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
```diff
@@ -78,6 +83,7 @@ public class RepositoryServiceDefault
     @Getter(onMethod_ = {@Override})
     final MetaModelContext metaModelContext;
 
+    private ThreadLocal<Map<Class, Boolean>> threadLocal = ThreadLocal.withInitial(HashMap::new);
```
danhaywood, line 86: give the `threadLocal` field a better name.

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
```diff
@@ -99,6 +105,27 @@ public <T> T detachedEntity(final @NonNull T entity) {
         return factoryService.detachedEntity(entity);
     }
 
+    @Override
+    public <T> T execInBulk(Callable<T> callable, Class<?>... classes) {
+        try {
+            Arrays.stream(classes).forEach(aClass -> setBulkMode(aClass, Boolean.TRUE));
+            return callable.call();
+        } catch (Exception e) {
```
danhaywood, line 113: I don't think this should swallow the exception... it should be propagated so that the transaction is aborted if there's a problem.

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
```diff
@@ -99,6 +105,27 @@ public <T> T detachedEntity(final @NonNull T entity) {
...
+            log.error("Error executing bulk", e);
+            return null;
+        } finally {
+            Arrays.stream(classes).forEach(aClass -> setBulkMode(aClass, Boolean.FALSE));
+        }
+    }
+
+    protected  <T extends Class> void setBulkMode(final T aClass, final Boolean bulkMode) {
```
danhaywood, line 121: Let's make this `private` visibility ; if it were subclassed then it could change the semantics completely.

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
```diff
@@ -99,6 +105,27 @@ public <T> T detachedEntity(final @NonNull T entity) {
         return factoryService.detachedEntity(entity);
     }
 
+    @Override
+    public <T> T execInBulk(Callable<T> callable, Class<?>... classes) {
```
danhaywood, line 109: A unit test for this method would be worthwhile, I think.

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
```diff
@@ -120,7 +147,9 @@ public <T> T persist(final T domainObject) {
     @Override
     public <T> T persistAndFlush(final T object) {
         persist(object);
-        transactionService.flushTransaction();
+        if (!threadLocal.get().getOrDefault(object.getClass(), Boolean.FALSE)) {
```
danhaywood, line 150: I'm wondering if this would work for JPA ... doesn't dynamic weaving create a subclass of the original entity?  
In which case, perhaps we can simplify - is there any reason to turn on bulk mode on a class-by-class basis ... why not simply turn it on "globally" (for the thread, that is)?

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
```diff
@@ -138,7 +167,9 @@ public void remove(final Object domainObject) {
     @Override
     public void removeAndFlush(final Object domainObject) {
         remove(domainObject);
-        transactionService.flushTransaction();
+        if (!threadLocal.get().getOrDefault(domainObject.getClass(), Boolean.FALSE)) {
```
danhaywood, line 170: same observation re JPA subtypes.

danhaywood (changes requested):
@opencirclesolutions-wove  - some comments for you to consider.  Also, as there is a config property that is set, I do think we need an integration test that verifies behaviour for different values.  Its default is 1, which means (if I follow) that we never trigger the new behaviour?

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Good point. The idea was to have more fine grain control over which classes you wanted to persist with or without a flush. But in bulk mode you want to have everything in one transaction. Everything else so have its own transaction or even thread.

core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
opencirclesolutions-wove: Didn’t know how strict your testing was. Apparently it’s quite strict.

api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
opencirclesolutions-wove: Makes sense

extensions/security/audittrail/applib/src/main/java/org/apache/causeway/extensions/audittrail/applib/dom/AuditTrailEntryRepositoryAbstract.java
opencirclesolutions-wove: Hmmm….

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Yes

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: That was the idea.

api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
opencirclesolutions-wove: Fixed

api/applib/src/main/java/org/apache/causeway/applib/services/repository/RepositoryService.java
opencirclesolutions-wove: Fixed

extensions/security/audittrail/applib/src/main/java/org/apache/causeway/extensions/audittrail/applib/dom/AuditTrailEntryRepository.java
opencirclesolutions-wove: Fixed

extensions/security/audittrail/applib/src/main/java/org/apache/causeway/extensions/audittrail/applib/dom/AuditTrailEntryRepositoryAbstract.java
opencirclesolutions-wove: You can call the super of an implemented interface even when it’s a default.

But this will do just fine:
return Can.ofCollection(repositoryService.execInBulk(() -> entityPropertyChanges.map(this::createFor).toList()));

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Fixed

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Fixed

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Then again why would you override it?

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Fixed

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Fixed

api/applib/src/main/java/org/apache/causeway/applib/services/publishing/spi/EntityPropertyChangeSubscriber.java
opencirclesolutions-wove: Fixed

core/config/src/main/java/org/apache/causeway/core/config/CausewayConfiguration.java
opencirclesolutions-wove: Added regression test with different causeway property.

persistence/commons/src/main/java/org/apache/causeway/persistence/commons/integration/repository/RepositoryServiceDefault.java
opencirclesolutions-wove: Changed an exisiting test to use bulk version.

PR #2465, opened 2024-06-04
CAUSEWAY-3753: adds additional validation of per params

danhaywood:
.. by extending Wicket's ScalarParameterModel

### DIFF
--- a/viewers/wicket/model/src/main/java/org/apache/causeway/viewer/wicket/model/models/ScalarParameterModel.java
+++ b/viewers/wicket/model/src/main/java/org/apache/causeway/viewer/wicket/model/models/ScalarParameterModel.java
@@ -18,18 +18,28 @@
  */
 package org.apache.causeway.viewer.wicket.model.models;
 
+import java.util.ArrayList;
+
 import org.apache.causeway.commons.collections.Can;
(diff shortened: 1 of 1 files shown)


viewers/wicket/model/src/main/java/org/apache/causeway/viewer/wicket/model/models/ScalarParameterModel.java
```diff
@@ -58,8 +68,80 @@ private ScalarParameterModel(
     }
 
     @Override
-    public String validate(final ManagedObject proposedValue) {
-        return getParameterNegotiationModel().getObservableParamValidation(getParameterIndex()).getValue();
+    public String validate(final ManagedObject proposedArg) {
+
+        val value = getParameterNegotiationModel().getObservableParamValidation(getParameterIndex()).getValue();
+        if (value != null) {
+            return null;
```
andi-huber, line 75: Is this logic flipped?

danhaywood: That is indeed a mistake, thanks.

PR #2550, opened 2024-07-09
Pre merge (3803)

danhaywood:
new config props for display of date and date/time


danhaywood:
@andi-huber  - could you review this?  I couldn't see another way to have Causeway render dates as "dd-MM-yyyy"

andi-huber (commented):
A fixed pattern here would disrespect any `TimeFormatStyle` (SHORT, MEDIUM, ...) fine-tuning (from the programming model) and also any end-user `Locale` settings from _SecMan_.

Hence, I'm not a fan of this change.

andi-huber:
Btw. if you are looking for ISO date format like `2024-12-31`, that would by provided by the combination of Locale `nds` (or e.g. `kw-GB`) and `FormatStyle.SHORT`. Which is supported by the framework.

danhaywood:
What the requirement is from our users is the same format that they had in v1, namely 24-05-2024.  Can this be done without this PR? I couldn't find a way.

I don't think iso format will be accepted.

PR #2766, opened 2024-10-31
CAUSEWAY-3815: Support for Actions in Collection View

andi-huber:
- [x] add meta-model method that list actions enabled for column rendering
- [x] render action drop-down menu to each row
- [ ] on action execution stay on page when action result matches element-type or is void, else redirect to blank page
- [ ] CSS polishing
- [ ] verify that we honor ARE_YOU_SURE semantics (danger indicator)
- [ ] verify dynamic visibility
- [ ] verify disabled indicator (and its tooltip)


andi-huber:
Lots of refactoring went into the Wicket Viewer, hence I'd like to merge into master and v3 before continuing work on this feature.
(I've added a config option that per default disables the feature for now.)

PR #3080, opened 2025-08-28
CAUSEWAY-3889: Support for object icon image embedding

andi-huber:
- [x] replace `iconName`, `iconData` and `faIconLayers` support methods with a single `icon` method
- [x] perhaps `iconName` may stay for backward compatibility (reinstate?)
- [x] UI Event for Icons needs an overhaul, only supports the `iconName` but should support the newer variants as well 
- [x] mignotes
- [ ] docs


api/applib/src/main/java/org/apache/causeway/applib/annotation/ObjectSupport.java
```diff
@@ -47,4 +50,31 @@
...
 public @interface ObjectSupport {
 
+    /**
+     * Rendering context for object icons.
+     * An object's icon is either rendered within the object detail page header (along its title)
+     * or within a table row.
+     */
+    public enum IconWhere {
+        OBJECT_HEADER,
+        TABLE_ROW //TODO also TREE_NODE and SELECT_DROPDOWN
```
andi-huber, line 60: Either rename, or add the other variants.

PR #3516, opened 2026-04-23
CAUSEWAY-3997: [Layout] In-memory Patching of Table Column Order

andi-huber:
- [x] API Design
- [x] - handle cases where domain-type not same as table's element-type
- [x] - handle wildcards / wildcard patching
- [x] Implement
- [ ] Test

Task-Url: https://issues.apache.org/jira/browse/CAUSEWAY-3997


api/applib/src/main/java/org/apache/causeway/applib/services/metamodel/MetaModelService.java
```diff
@@ -166,13 +166,61 @@ enum Mode {
...
+     *
+     * <p> Where columns returned are those that are in principle available. Also those are returned in no particular order.
+     *
+     * <p> If domainType is null returns an empty {@link Stream}.
+     *
+     * <p> Otherwise, if memberIdentifier is null returns <i>Columns</i> for the standalone table of element-type == domainType.
+     *
+     * @since 4.0
+     */
+    Stream<Identifier> streamAvailableAssociationsForColumnRendering(@Nullable Class<?> domainType, @Nullable Identifier memberIdentifier);
```
andi-huber: Not totally happy with that yet, as 
1. distinction between PARENTED and STANDALONE could be more explicit (separate methods)
2. column-order-txt feature supports identifier wildcards, which could be reflected by leaving the memberId empty here

(same applies to the next method below)

api/applib/src/main/java/org/apache/causeway/applib/services/columnorder/Object_patchColumnOrder.java
```diff
@@ -0,0 +1,123 @@
...
+
+	public static class ActionDomainEvent
+	extends org.apache.causeway.applib.CausewayModuleApplib.ActionDomainEvent<Object_patchColumnOrder> {}
+
+	@Inject MetaModelService metaModelService;
+	@Inject ApplicationFeatureRepository applicationFeatureRepository;
+
+    private final Object mixee;
+
+    @MemberSupport public Object act(
```
andi-huber, line 88: general open question: how to deal with cases, where the concrete domain-object is an instance of the table's element type, but not the exact same type?

we'd need another parameter, that reflects type inheritance and provides a means for users to make their element-type choice

andi-huber: redesign allows wildcards for PARENTED and type selection for STANDALONE

andi-huber:
closing ... not yet tested (de-scoped): table export, standalone tables

core/metamodel/src/main/java/org/apache/causeway/core/metamodel/spec/impl/_MembersAsColumns.java
```diff
@@ -18,99 +18,113 @@
...
+    public Stream<ObjectAction> streamActionsForColumnRendering(
             final ObjectSpecification elementType,
             final Identifier memberIdentifier) {
-        if(elementType.isValue()) return Stream.empty();
+        if(elementType.isValue())
+			return Stream.empty();
 
         return elementType.streamRuntimeActions(MixedIn.INCLUDED)
-            .filter(ObjectAction.Predicates.visibleAccordingToHiddenFacet(WhereContexts.collectionVariant(memberIdentifier)))
+            .filter(ObjectAction.Predicates.visibleAccordingToHiddenFacet(Where.STANDALONE_TABLES))
```
andi-huber, line 69: needs fixing: using STANDALONE always here is wrong

also we cannot determine STANDALONE vs PARENTED from just the Identifier as we did previously; I believe we need the information explicitly

PR #3552, opened 2026-05-12
Add explicit workflow permissions for weekly build job

arpitjain099:
## Summary
- Add an explicit `permissions` block to `.github/workflows/ci-build-artifacts-push-maven.yml`.
- Scope token access to `contents: read` and `packages: write` for the deploy workflow.

## Why
Declaring explicit workflow token permissions follows least-privilege defaults and makes required access clear for scheduled automation.


andi-huber:
Found related docs here https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

However, not going to merge, as needs testing and the targeted workflow is currently disabled anyway. We'll also probably not enable any time soon.

But regardless, thanks for the hint! Might be useful for other workflows.

PR #3566, opened 2026-05-20
CAUSEWAY-3997: [v2] Backport of Column Order Patching from v4 to v2

andi-huber:
- [x] Listing
- [x] ListingTest
- [x] ApplicationFeatureRepository
- [x] MetaModelService
- [x] ColumnOrderPatchingFacet
- [x] Object_patchColumnOrder Mixin
- [x] Test using Demo v2

see https://github.com/apache/causeway/pull/3516

### DIFF
--- a/core/metamodel/src/main/java/org/apache/causeway/core/metamodel/specloader/specimpl/dflt/_MembersAsColumns.java
+++ b/core/metamodel/src/main/java/org/apache/causeway/core/metamodel/specloader/specimpl/dflt/_MembersAsColumns.java
@@ -29,7 +29,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Stream;
 
-import org.apache.causeway.applib.Identifier;
 import org.apache.causeway.applib.annotation.Where;
 import org.apache.causeway.applib.layout.component.PropertyLayoutData;
 import org.apache.causeway.applib.services.tablecol.TableColumnOrderService;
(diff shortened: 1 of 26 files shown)


core/metamodel/src/main/java/org/apache/causeway/core/metamodel/specloader/specimpl/dflt/_MembersAsColumns.java
```diff
@@ -18,102 +18,118 @@
...
+    public Stream<ObjectAction> streamActionsForColumnRendering(
             final ObjectSpecification elementType,
             final Identifier memberIdentifier) {
-        if(elementType.isValue()) return Stream.empty();
+        if(elementType.isValue())
+			return Stream.empty();
 
         return elementType.streamRuntimeActions(MixedIn.INCLUDED)
-            .filter(ObjectAction.Predicates.visibleAccordingToHiddenFacet(WhereContexts.collectionVariant(memberIdentifier)))
+            .filter(ObjectAction.Predicates.visibleAccordingToHiddenFacet(Where.STANDALONE_TABLES))
```
andi-huber: needs fixing: using STANDALONE always here is wrong

also we cannot determine STANDALONE vs PARENTED from just the `Identifier` as we did previously; I believe we need the information explicitly

andi-huber:
Tests succeed on local build. Not sure what's happening with those Github runners at the moment.

